PR #39, opened 2023-12-12
Support Python 3.8

mariosasko:
This PR makes the following changes to support Python 3.8:
* replace the `match-case` statement with `if-else`
* use the `{**a, **b}` syntax to merge dicts instead of `a | b`
* use `typing.Union` instead of the `|` operator and the `typing` versions of the built-in types (`List`, `Dict`, etc.) instead of subscripting them.
* installs the backported `importlib.resources` package on Python<3.9 to read `assets`
* modifies the required NumPy version in `setup.py` to support Python 3.8

(Python 3.7 came to the "end of life" 6 months ago, so dropping support for it should be fine)

### DIFF
--- a/src/datatrove/executor/base.py
+++ b/src/datatrove/executor/base.py
@@ -30,7 +30,7 @@ class PipelineExecutor(ABC):
             skip_completed: whether to skip tasks that were completed in
                 previous runs. default: True
         """
-        self.pipeline: list[PipelineStep | Callable] = pipeline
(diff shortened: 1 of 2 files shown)


chschroeder:
@guipenedo Just wanted to try out datatrove again after some time and realized it needs Python 3.10. Which Python 3.10 "very nice" features are really needed here? 

This is really a reason to put datatrove aside.

PR #56, opened 2024-01-21
Use re search instead of findall

fierzdev:
In the regex docstring is written "filters if regex find at least one match". This means that findall is an overkill, as we are not actually interested in all occurrences. For a few datapoints this might not be noticeable and depending on the regex pattern, it will also not make a difference. But in some cases this could lead to a speed up of the pipeline, so I guess would be worth changing.


guipenedo:
Well spotted, thanks!

Issue #54, opened 2024-01-19
replace setup.py with pyproject.toml


baggiponte:
Hey there, glad to see there's something knew cooking with HF! I hope this project gets traction, it seems very interesting.

I noticed your setup.py is really straightforward. Its contents could simply be put inside the `pyproject.toml`: unless you are compiling C code, you won't likely need a `setup.py` to define the metadata of your package.

The biggest reason to use this format is that it's the standard that Python devs are [strongly recommended to adopt](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#is-pyproject-toml-mandatory). Practically speaking, `pyproject.toml` is a plain text file so it can be used to expose metadata about the project in PyPI, while `setup.py` needs to be run to establish them (say you have an if statement somewhere: pip does not know it by hand). This is especially useful when building wheels locally, though it should not happen on most platforms.

Using `pyproject.toml` also means that every package manager that complies with PEP518 (basically, everything but poetry) can be used to develop on this project.

Hope I was clear enough. If you are interested I'd be glad to support you 🤗

guipenedo:
Would you be interested in making this change in a PR?

baggiponte:
Definitely! :)

PR #59, opened 2024-01-22
build: replace setup.py with pyproject.toml

baggiponte:
Addresses #54. The library builds correctly; I also ran the tests and I get some failures because some resources from nltk are not installed, e.g.:

```
>       raise LookupError(resource_not_found)
E       LookupError:
E       **********************************************************************
E         Resource punkt not found.
E         Please use the NLTK Downloader to obtain the resource:
E
E         >>> import nltk
E         >>> nltk.download('punkt')
E
E         For more information see: https://www.nltk.org/data.html
E
E         Attempted to load tokenizers/punkt/PY3/english.pickle
E
E         Searched in:
E           - '/Users/luca/nltk_data'
E           - '/Users/luca/dev/contrib/datatrove/venv/nltk_data'
E           - '/Users/luca/dev/contrib/datatrove/venv/share/nltk_data'
E           - '/Users/luca/dev/contrib/datatrove/venv/lib/nltk_data'
E           - '/usr/share/nltk_data'
E           - '/usr/local/share/nltk_data'
E           - '/usr/lib/nltk_data'
E           - '/usr/local/lib/nltk_data'
E           - ''
E       **********************************************************************
```

I have a fix for this but I think it should be another PR.

Notes: the following `setup.py` sections are not mapped to their equivalent in `pyproject.toml` because setuptools has package discovery:

```
setup(
    ...
    packages=find_packages("src"),
    package_dir={"": "src"},
)
```

While `include_package_data=True,` is true by default.

One thing: the `project.scripts` you create wil be available under the names `check_dataset` etc (this is regardless of whether you want to use pyproject or not). If you want them as subcommands of a `datatrove` lib, you should use `click` or `typer` to write a CLI.


pyproject.toml
```diff
@@ -1,16 +1,120 @@
...
+datatrove = ["assets/*"]
+
 [tool.ruff]
-# Ignored rules:
-#   "E501" -> line length violation
-#   "C901" -> `function_name` is too complex
-ignore = ["C901", "E501"]
-select = ["C", "E", "F", "I", "W"]
+ignore = [
+  "C901", # line length violation
```
guipenedo, line 98: the comments ended up switched I believe. C901 is "function name is too complex"

guipenedo:
Tests seem to work (you need to run `python -m nltk.downloader punkt` before running the tests)

baggiponte:
> Tests seem to work (you need to run `python -m nltk.downloader punkt` before running the tests). We just have to update the code quality action as that currently checks setup.py

yes! I would suggest/implement adding a try-catch blog that does that for you.

pyproject.toml
```diff
@@ -1,16 +1,120 @@
+[build-system]
+requires = ["setuptools"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "datatrove"
+version = "1.0.1.dev0"  # expected format is one of x.y.z.dev0, or x.y.z.rc1 or x.y.z (no to dashes, yes to dots)
+description = "HuggingFace library to process and filter large amounts of webdata"
```
guipenedo, line 8: is it not possible to also include the `long_description` (the readme) as before?

guipenedo (changes requested):
Could you also remove `setup.py` from `.github/workflows/ci.yml`'s ruff commands?

pyproject.toml
baggiponte: Solved in the last commit I pushed, thanks for spotting that :)

pyproject.toml
baggiponte: The pyproject [spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/) follows PEP621 and does not include a `long_description` field, but the readme can be included and specified in the `readme` key of the `project` table. Should be line 9 in this doc.

guipenedo (approved):

PR #61, opened 2024-01-22
Add goose3 extractor

### LINKED ISSUE #55: License
Hi,
I noticed this project uses Trafilatura, which is GPL licensed. Would you consider switching to an alternate library such as goose3 to avoid GPL restrictions?
Thanks!

fakerybakery:
Fixes #55

### DIFF
--- a/src/datatrove/pipeline/extractors/goose3.py
+++ b/src/datatrove/pipeline/extractors/goose3.py
@@ -1,0 +1,45 @@
+// Goose3 extractor for DataTrove - extract text with goose3
+// mrfakename (www.mrfake.name)
+// Apache License 2.0
+from .base import BaseExtractor
+
+
+class Goose3(BaseExtractor):
+    """Goose3 extractor, it uses https://github.com/goose3/goose3"""
+
+    name = "⛏ Goose3"
+    _requires_dependencies = ["goose3"]
+
+    def __init__(
+        self,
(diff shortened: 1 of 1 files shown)


guipenedo (changes requested):
Thank you for the PR! Currently the following minimal example doesn't seem to run:
```
ARTICLE_HTML = "<html><body><article><p>Hello World!</p></article></body></html>"
extractor = Goose3()
print(extractor.extract(ARTICLE_HTML)) # should be "Hello World!"
```
I have added some comments that when addressed should allow the code to run.
Also, can you please run the linter and fix the issues so that we have consistent style:
`ruff src`
Another note: you can just add more commits and the PR will update, no need to delete and re-create it

src/datatrove/pipeline/extractors/goose3.py
```diff
@@ -0,0 +1,45 @@
+# Goose3 extractor for DataTrove - extract text with goose3
+# mrfakename (www.mrfake.name)
```
guipenedo, line 2: usually as a principle HuggingFace's libraries do not include individual contributor's copyright notices
Your contribution would be indicated on a changelog (and obviously live forever on the commit history) but not so much on the header of files

fakerybakery:
Sorry about that, Trafilatura is probably better

PR #71, opened 2024-01-30
Zero Division Error Fix

NicholasLindner:
Add an if statement to check if len(freq) is 0. If it is, 0 is returned. This prevents zero division error.

### DIFF
--- a/src/datatrove/pipeline/filters/unigram_log_probs.py
+++ b/src/datatrove/pipeline/filters/unigram_log_probs.py
@@ -56,9 +56,7 @@ class UnigramLogProbFilter(BaseFilter):
         from nltk.tokenize import word_tokenize
 
         words = word_tokenize(doc.text)
-        freqs = [
-            self.unigram_frequencies.get(word.lower(), 0) for word in words
-        ]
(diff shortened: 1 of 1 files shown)


guipenedo:
Actually thinking about it a bit more, if a word is not present in the frequencies dictionary, then we should assume its frequency is 0 (and have that value influence the mean) instead of ignoring it. We can keep the check you added for the case when a sample has no words at all, but it would be good to also change the line that gets all the frequencies by removing the check that ensures it is in the dictionary and adding `0` as a default argument on `get`

NicholasLindner:
Ok I think I did it :)

guipenedo:
On second thought this won't work as we are taking the log in the end. It should be a value different from 0, possibly 1/10 of the smallest value in the frequencies table. Will take a better look tomorrow.
Sorry about that

NicholasLindner:
@guipenedo Hi! Did you get a chance to look over it? Is it fine as it is or do you think it needs another change?

guipenedo:
Yeah let's change the 0 to `1e-9`. The lowest frequency is around 2.2e-8, so this seems like a decent enough value for an "unknown" unigram.

guipenedo:
Thanks for the help @NicholasLindner

PR #73, opened 2024-01-31
Update exact_substrings.py

jordane95:
Fix indent bug

### DIFF
--- a/src/datatrove/pipeline/dedup/exact_substrings.py
+++ b/src/datatrove/pipeline/dedup/exact_substrings.py
@@ -315,7 +315,7 @@ class DedupReader(JsonlReader):
         # loads the sequence file from stage 1, the size file from stage 1 and the bytearange file.
         sequence_file, size_file = self.get_all_files(rank=self.rank, world_size=world_size)
         if not self.dup_ranges:
-            yield None
+            return
         # data is given only during tests.
         if not data:
             data = self.read_files_shard(self.data_folder.get_shard(self.rank, world_size))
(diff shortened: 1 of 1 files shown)


jordane95:
Hi @guipenedo 

I find this line throws error when the data is empty, i is not defined.

After second thought, this isn't a good fix since it is intended to check it only once at the end of the iteration?

guipenedo:
Thanks for the PR! Indeed, it should be outside the for loop. Maybe just add a `i = 0` above the for loop?

jordane95:
Exactly! But i should be -1 in this case to pass the assertion of doc_lens

guipenedo:
You're right yeah. Could you then please revert your original change and remove the indent for the 2 asserts?

jordane95:
Done! Actually the previous one was already outside the `for` loop but inside the `with` context, should make on difference

src/datatrove/pipeline/dedup/exact_substrings.py
```diff
@@ -313,9 +314,11 @@ def run(self, data: DocumentsPipeline = None, rank: int = 0, world_size: int = 1
         self.rank = rank
         # loads the sequence file from stage 1, the size file from stage 1 and the bytearange file.
         sequence_file, size_file = self.get_all_files(rank=self.rank, world_size=world_size)
+        if not self.dup_ranges:
+            yield None
```
guipenedo, line 318: maybe just `return` instead?

guipenedo:
I admit it has been a while since I last ran this block, thanks for the changes. Is it currently running well with the new changes? I think the tests for the last block are lacking a bit

src/datatrove/pipeline/dedup/exact_substrings.py
jordane95: Yeah, `return` works.

jordane95:
Not yet. I will look into it and make some changes soon

jordane95:
Actually this issue is also fixed by early existing.

> does this also fix #76 ?

guipenedo:
Perfect! Thanks a lot for all the work, merged

PR #84, opened 2024-02-03
Add prefix to slurm logs

loubnabnl:
option to add a prefix to slurm logs (could be useful when tokenizing multiple sources in the same `logging_dir`)


guipenedo:
I would generally advise against running different jobs in the same directory, as the completion folder, the logs and stats would get overwritten between jobs. It's usually cleaner to change the `logging_dir` directly by appending the prefix to it

loubnabnl:
Ok that makes sense, closing the PR

PR #86, opened 2024-02-08
Changed fsx default filepath for logging output to user's home

Anacheron51:
Changed fsx default filepath for logging output to user's home. Default file path uses a personal fsx instance, which breaks if the user does not have access to it.


Anacheron51:
> Hi, the PR seems to be empty, I don't think it registered your changes

Added it!

guipenedo:
ty!

PR #90, opened 2024-02-12
Adding doc strings + adding a faster tokenized doc merger

thomwolf:


src/datatrove/pipeline/tokens/merger.py
```diff
@@ -10,9 +11,225 @@
 from datatrove.pipeline.tokens.tokenizer import TokenizedFile
 
 
+class FileTokenizerMerger(PipelineStep):
```
guipenedo, line 24: Have thoughts on this, messaged you on slack

src/datatrove/pipeline/tokens/tokenizer.py
```diff
@@ -197,14 +204,35 @@ def save_final_metadata(self, tokenizer_name: str | None = None, token_count: in
...
+    """
     name = "✍️ Writer"
     type = "🔢 - TOKENIZER"
     _requires_dependencies = ["tokenizers"]
 
     def __init__(
         self,
         output_folder: DataFolderLike,
-        local_working_dir: str | None = None,
+        local_working_dir: Optional[Tuple[DataFolderLike, str]],
```
guipenedo, line 235: Why `Optional`? `|` is from Python 3.10 which is our minimum version. Also, it should be a `DataFolderLike` and not a tuple

src/datatrove/pipeline/tokens/tokenizer.py
```diff
@@ -222,6 +250,8 @@ def __init__(
         self.local_working_dir = get_datafolder(local_working_dir) if local_working_dir else None
         if self.local_working_dir and not isinstance(self.local_working_dir.fs, LocalFileSystem):
             raise ValueError("local_working_dir must be a local path")
+        if self.local_working_dir is None:
```
guipenedo, line 253: The check should be:
If local_working_dir is None AND shuffle is True AND output_dir is not a local directory

(view line above for this last one)

guipenedo (changes requested):
Thanks for the new docstrings and the fix on the urlstats. I'm okay with moving the docstrings from __init__ up to the top of the classes.

There are some changes that will break things and I don't see a reason for (`fork`, the logging one, removing `tolist` at one point).
For the new tokens merger I messaged you directly.
The tests are also broken (code quality etc)

guipenedo (changes requested):

guipenedo (approved):

PR #96, opened 2024-02-20
Add adapter in DedupReader

jordane95:
I think we should add an adapter here so we won't be bother with saving an intermediate json file if we are working with a different format of jsonl data


guipenedo:
to be fair I'd be more in favour of changing DedupReader to not subclass a reader and just get its data normally from the `run` parameters (so that you can use whatever writer/reader combo you want to save intermediate data)

guipenedo:
lmk if this would also work for you: https://github.com/huggingface/datatrove/pull/98/

PR #98, opened 2024-02-20
Decoupled reading logic from DedupReader

guipenedo:
- decouples document reading logic from the stage 3 exactsubstring block
- renamed blocks to include ES prefix
- renamed `DedupReader` to `ESRangeRemover`
- updated tests
- updated example


jordane95:
Yeah, this is better

PR #123, opened 2024-03-13
Fix `c4_qualiter_filter` curly brackets detection error

koalazf99:
The original implementation use `if "{" or "}" in doc.text:` to filter the curly bracket
which actually will always produce `True`, so all document will be filtered.

Quick fix: use `if "{" in doc.text or "}" in doc.text:` instead.


guipenedo:
Thanks for contributing! I am actually in the process of rewriting this filter and have already fixed this issue :)
I will tag you in the PR when it is ready so you can take a look

koalazf99:
Wow that's very cool, can't wait to see!😊 

Again, thanks for your effort, really good project for me.

guipenedo:
see #124

PR #126, opened 2024-03-15
Bug fix: when file is empty

jordane95:
This may raise error when running `seek_to_start` func with file size = 0

### DIFF
--- a/src/datatrove/pipeline/dedup/minhash.py
+++ b/src/datatrove/pipeline/dedup/minhash.py
@@ -152,6 +152,8 @@ def read_sigs(
         index_file: is index file
     """
     with file as f:
+        if f.size == 0:
+            yield None
         seek_to_start(f, min_hash, config, index_file)
         last = None
         for data in read_tuples_from_file(


guipenedo:
Thanks for the PR! Indeed I think this fixes the issue but maybe it's cleaner to replace the `yield None` with a simple `return`?

src/datatrove/pipeline/dedup/minhash.py
```diff
@@ -152,6 +152,8 @@ def read_sigs(
         index_file: is index file
     """
     with file as f:
+        if f.size == 0:
+            yield None
```
guipenedo, line 156:
```suggestion
            return
```

jordane95: I don't think we can simply return None here since the value of this function is used as an iterator in latter code

https://github.com/huggingface/datatrove/blob/a98aafd2f3fe3dab3addd8ad9483338e92494938/src/datatrove/pipeline/dedup/minhash.py#L386-L402

guipenedo: It should still work, as long as you have a `yield` inside the function body it should be treated as a generator even if the yield is never reached

jordane95: Magic

Issue #121, opened 2024-03-12
Potential issues in substring dedup


jordane95:
Hi @guipenedo , I used your substring dedup script to perform deduplication on a dump of cc and did some manual inspection. I find that some resulting duplicates a bit strange.

For example,

```
{
  "id": "sha1:222JEVYQHVGRTKDSUSPOJEVUETA5AEO6",
  "data": [
    {
      "meta": {
        "bucket": "head",
        "date_download": "2021-01-26T19:40:47Z",
        "language": "en",
        "language_score": 0.86,
        "perplexity": 286.2,
        "source_domain": "microbewiki.kenyon.edu",
        "title": "Difference between revisions of \"Streptococcus salivarius\" - microbewiki",
        "url": "https://microbewiki.kenyon.edu/index.php?title=Streptococcus_salivarius&diff=132493"
      },
[...]

PR #132, opened 2024-03-22
Fix substring dedup range

jordane95:
We should avoid byte bifurcation in bytes sequence of int16 since the google script doesn't account for this


jordane95:
This should partially solve #121

PR #136, opened 2024-03-23
Implement LanguageIdStats Feature and Metadata Check in language_filter

justHungryMan:
For processes handling multi-language, it was necessary to consider language ID stats, hence the addition of the following features.

- LanguageIdStats Feature: A new feature designed to enhance language identification capabilities. LanguageIdStats provides statistical insights into language usage, aiding in more accurate and nuanced language detection and analysis.

- Metadata Check in language_filter: To complement the existing filtering mechanisms, a metadata checking feature has been integrated into the language_filter module. This allows for more refined filtering based on specific metadata criteria, improving the overall precision of language selection.

Additionally, a point for consideration is where to store the fastText model. Since the same model is used in the language_filter, it has been designated to the same location, but this matter requires further discussion.

### DIFF
--- a/examples/stats_language_id.py
+++ b/examples/stats_language_id.py
@@ -1,0 +1,19 @@
+from datatrove.executor.base import PipelineExecutor
+from datatrove.executor.local import LocalPipelineExecutor
+from datatrove.pipeline.readers import HuggingFaceDatasetReader
+from datatrove.pipeline.stats import LanguageIdStats
+from datatrove.pipeline.writers.jsonl import JsonlWriter
+
+
+def run_step():
+    pipeline_1 = [
+        HuggingFaceDatasetReader(
+            "truthful_qa", dataset_options={"name": "generation", "split": "validation"}, text_key="question"
+        ),
+        LanguageIdStats(),
+        JsonlWriter("truthful_qa_with_language_id", compression=None),
+    ]
+
+    executor_1: PipelineExecutor = LocalPipelineExecutor(pipeline=pipeline_1, workers=1, tasks=1)
+
+    executor_1.run()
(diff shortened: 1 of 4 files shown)


guipenedo:
I'm not sure we need a new block for this, maybe we could simply add an "all" option to keep all the languages in the current filter and also add the line that keeps track of the number of samples per language to the current filter?

justHungryMan:
I think language_filter should use to its filtering function. Functionally, I think you could simply add an all option like you said, but semantically, I think it should be separated the two functions, for filtering and for statistics.

justHungryMan:
Hi @guipenedo 

Could we consider creating this functionality directly, or would it be better to use the existing language_filter with a threshold set to 0 to achieve this indirectly? In terms of dealing with multi-lingual, it is necessary to get the "language" field.

Looking forward to your thoughts and guidance.

guipenedo:
Hi,
sorry for the late reply and for leaving the PR a bit abandoned. 
> In terms of dealing with multi-lingual, it is necessary to get the "language" field.

Basically the approach we are thinking of for multilinguality is to have users explicitly specify the language that should be used for word tokenization, filtering thresholds etc, rather than fetch the language from the metadata field.
Our thinking for this decision was:
- less error prone, as you explicitly set the language
- people might want to run this on mixed language data, for instance if you have a document with the majority of the text in english (the value in the metadata) and the rest in some minority language

So a possible setup would be to use the filter to sort the data into different folders (one for each language) as a first step, followed by running a different pipeline for each one

> Could we consider creating this functionality directly, or would it be better to use the existing language_filter with a threshold set to 0 to achieve this indirectly?

We could still have a standalone block for this, though as I said in the beginning I am not sure it really is needed as I could easily see a `label_only` option or similar getting added that would basically keep all the data and only add the metadata label

justHungryMan:
I wanted to process WARC data from multiple languages at once. To do that, I needed the language field for tokenizing.
I also agree with your opinion for duplicating issue in my code and also processing with different folders (one for each language)

So, can I add `label_only` feature in `lanugage_filter.py` and PR?
I think this PR can be closed.

guipenedo:
> So, can I add label_only feature in lanugage_filter.py and PR?

Sounds great!

PR #139, opened 2024-03-26
Fix: Handle Non-dict Objects in to_dict Without Errors

justHungryMan:
Ran into a bug where our to_dict method in stats.py would crash if it hit an int or something without a to_dict method. 
### error msg
``` error msg
/lib/python3.11/site-packages/datatrove/utils/stats.py", line 52, in to_dict
    t[a] = b.to_dict()
           ^^^^^^^^^
AttributeError: 'int' object has no attribute 'to_dict'
```

### sample data
```sample data
[
    {
        "name": "\ud83d\udcd6 - READER: \ud83d\udcd2 Parquet",
        "time_stats": {
            "total": 13.128175735473633,
            "mean": 13.128175735473633,
            "unit": "batch",
            "total_human": "13 seconds",
            "mean_human": "13 seconds and 128.18 milliseconds",
            "std_dev_human": "0 milliseconds",
            "min_human": "13 seconds and 128.18 milliseconds",
            "max_human": "13 seconds and 128.18 milliseconds"
        },
        "stats": {
            "input_files": 1,  # <— THIS CASE
            "doc_len": {
                "total": 177830,
                "n": 49,
                "mean": 3629.183673469389,
                "variance": 15571471.153061228,
                "std_dev": 3946.070343146613,
                "min": 39,
                "max": 18169
            },
            "documents": {
                "total": 48,
                "mean": 48.0,
                "unit": "input_file"
            }
        }
    },
…
]
```

Crashes with an AttributeError when it hits basic data types (like our friend, the integer).

The Fix:
- Added a quick check to only call to_dict() if the object actually has it. If not, just return the object as is.
- Now it handles int, floats and whatever else doesn't play nice with to_dict.

### DIFF
--- a/src/datatrove/utils/stats.py
+++ b/src/datatrove/utils/stats.py
@@ -45,7 +45,7 @@ class MetricStatsDict(defaultdict):
         return ", ".join(f"{key}: {stats}" for key, stats in self.items())
 
     def to_dict(self):
-        return {a: b.to_dict() for a, b in self.items()}
+        return {a: (b.to_dict() if hasattr(b, "to_dict") else b) for a, b in self.items()}
 
 
 class Stats:


hynky1999:
Thanks for the PR!
While I understand your idea, the thing is that the class that owns this `MetricStatsDict`, which is in fact `defaultdict(MetricStats)`. This means, that we expect that each value is of type `MetricStats` which do have `to_dict`. There are also other methods that work with this assumption.

The easy fix for you should be that instead of `int` as as your value for `input_files`, you should use `MetricStats`.

justHungryMan:
Hi @hynky1999,

It appears there's a bit of confusion regarding the issue. The error manifests during the execution of a Slurm job, specifically when merging the statistics generated by individual Slurm tasks. The example data I shared illustrates the scenario where the problem arises (indicated by the comment "#<-- THIS CASE").

If your assessment holds, it suggests that the discrepancy lies within the process that merges these statistics, not necessarily in the data structure itself.

The essence of my pull request is to introduce a safeguard against the bug we encountered amidst debugging efforts with the current Slurm setup. This proactive measure is aimed at ensuring our to_dict method in stats.py can gracefully handle situations where it encounters non-dict objects, like integers, by checking for the presence of a to_dict method before invoking it.

guipenedo:
Hi @justHungryMan, indeed you are quite right, to simplify the Json structure stats and do not have any variability/mean etc are saved as simple ints. But in our `from_dict` method there is a check to correctly load these integers, so I'm not sure why the error you are experiencing was triggered.
Could you please share the pipeline code (including executor) you used so that we can reproduce the issue?

justHungryMan:
I'll report back with the pipeline reproduction code soon. Currently tied up with some tasks, but I'll share it as soon as possible. Apologies for the delay.

justHungryMan:
Based on the PR discussion, I believe I am currently modifying the datatrove code due to requirements in our Slurm environment. Specifically, the use of sqsh files (docker images) is essential, which led me to make modifications to the internal code. The need for these changes arose from errors encountered during the merging of result files, making me to submit this PR. Given our setup, which involves thousands of tasks and hundreds of workers, it becomes challenging to provide a reproduction command. Additionally, it's impossible for us to replicate this issue using venv or conda, as their use is not feasible in our environment.

When I examined the results from datatrove locally, I confirmed that the `input_files` part still enters as an integer type. If the `merge_stats` function is not the one modifying this, then it appears my report might only be reproducible in our unique environment. However, if the `merge_stats` function does indeed interact with the `input_files` section, then it seems the defensive code I proposed is necessary.

Do you have any thoughts on this matter?

guipenedo:
Indeed it seems to be a hard to reproduce issue. In any case I think the change you added does not create any harm so I will merge it as is. Thanks!

PR #144, opened 2024-04-01
add requeue and QOS slurm options

marianna13:
1. Not all clusters support QOS and requeue options so we make them optional. 
2. `launch_pickled_pipeline` might not work all the time (e.g. if you use a local fork of datatrove)


guipenedo:
Hi! Thanks for the changes, they look great!
Was just unsure as to why having a local fork would make the launch pipeline command not work. On my system installing from local source (`pip install . -e`) adds the command to the environment without issue, does this not happen for you?

guipenedo:
Thank you for the PR, I made a small change and fixed the style. Merged!

PR #147, opened 2024-04-02
[WIP] Multi-Lingual Tokenization

beme248:


pyproject.toml
```diff
@@ -54,7 +54,15 @@ processing = [
     "tldextract",
     "trafilatura>=1.8.0",
     "tokenizers",
-    "ftfy"
+    "ftfy",
+    "stanza",
```
beme248, line 58: I suggest to keep only necessary dependency for Korean and also only code to support Korean besides English for now. It will make it easier  to review the PR and we can easily add more Languages once the structure is agreed on.

src/datatrove/tools/word_tokenizers.py
```diff
@@ -0,0 +1,197 @@
...
+    Languages.western_frisian: NLTKTokenizer("dutch"),  # Proxy
+    Languages.breton: StanzaWordTokenizer("cy"),  # Proxy
+    Languages.malagasy: NLTKTokenizer("english"),  # Proxy
+    Languages.yiddish: SpaCyTokenizer("he"),  # Proxy
+    Languages.somali: NLTKTokenizer("english"),  # Proxy
+    Languages.turkmen: NLTKTokenizer("turkish"),  # Proxy
+}
+
+
+def get_word_tokenizer(language: str):
```
beme248, line 193: This code is currently `dead code`. Can we add a test that guarantees that the behavior for English does not change and then replace:

```
        from nltk.tokenize import word_tokenize

        words = word_tokenize(doc.text)
   ```
  with automatic language specific tokenisation?

Comment (vsabolcec) [reply src/datatrove/tools/word_tokenizers.py:45]: Currently, the tests in [`tests/pipeline/test_filters.py`](https://github.com/huggingface/datatrove/blob/afadc8f20f7666b86743a9554d231c0177b2f39f/tests/pipeline/test_filters.py#L37) don't include language metadata. Including automatic language specific tokenization in the filters makes the tests fail. Should we also modify the tests to support the change?

File: pyproject.toml [hunk 3]
```diff
@@ -54,7 +54,8 @@ processing = [
     "tldextract",
     "trafilatura>=1.8.0",
     "tokenizers",
-    "ftfy"
+    "ftfy",
+    "spacy",
```
vsabolcec, line 58: SpaCy dependency for the Korean language tokenizer is currently in the "processing" group. However, as more tokenizers are added in the future, we would like to minimize the number of dependencies that the user has to download if they do not need to process all languages.

src/datatrove/pipeline/filters/gopher_quality_filter.py
```diff
@@ -66,10 +67,10 @@ def filter(self, doc: Document) -> bool | tuple[bool, str]:
         Returns: False if sample.text does not pass any of the the heuristic tests
 
         """
-        from nltk.tokenize import word_tokenize
-
         text = doc.text
-        words = word_tokenize(text)  # TODO we should use language id filter
+        language = doc.metadata.get("language", "en")
```
vsabolcec, line 71: To ensure backward compatibility, we add "en" as the default language if no language metadata is provided. Do we want to handle backward compatibility in another way?

src/datatrove/tools/word_tokenizers.py
```diff
@@ -0,0 +1,49 @@
...
+            token.text
+            for token in self.tokenizer(text, disable=["parser", "tagger", "ner"])
+            if len(token.text.strip()) > 0
+        ]
+
+
+WORD_TOKENIZERS: dict[str, WordTokenizer] = {
+    Languages.english: NLTKTokenizer("english"),
+    Languages.korean: SpaCyTokenizer("ko", {"nlp": {"tokenizer": {"@tokenizers": "spacy.Tokenizer.v1"}}}),
+}
```
vsabolcec, line 42: Currently, tokenizers are created when the module is imported. Ideally, we would want to lazy-load only the tokenizers that the user (explicitly) or language metadata (implicitly) requests.

tests/pipeline/test_word_tokenizers.py
```diff
@@ -0,0 +1,31 @@
...
+            is_stripped = [token == token.strip() for token in tokens]
+            assert all(is_stripped), f"'{language}' tokens contain whitespaces"
+
+    def test_english_tokenizer(self):
+        nltk_words = word_tokenize(SAMPLE_TEXT, language="english")
+
+        tokenizer = get_word_tokenizer(Languages.english)
+        tokenizer_words = tokenizer.tokenize(SAMPLE_TEXT)
+
+        self.assertEqual(nltk_words, tokenizer_words, "NLTK tokenizer and multilingual tokenizer differ")
```
vsabolcec, line 31: This test ensures backwards compatibility for the English tokenizer only. Do we also want to add some integration tests to ensure that the change is working correctly?

src/datatrove/pipeline/filters/gopher_quality_filter.py
```diff
@@ -66,10 +67,10 @@ def filter(self, doc: Document) -> bool | tuple[bool, str]:
         Returns: False if sample.text does not pass any of the the heuristic tests
 
         """
-        from nltk.tokenize import word_tokenize
-
         text = doc.text
-        words = word_tokenize(text)  # TODO we should use language id filter
+        language = doc.metadata["language"]
+        tokenizer = get_word_tokenizer(language)
+        words = tokenizer.tokenize(text)
```
vsabolcec, line 73: This is the proposed change that would be applied to all places where tokenization is performed. How do we want to change the API to provide backwards compatibility and handle cases where language is not provided in the document metadata?

guipenedo:
@vsabolcec @beme248 please let me know when you are ready for me to review

src/datatrove/pipeline/filters/gopher_quality_filter.py
```diff
@@ -66,10 +69,9 @@ def filter(self, doc: Document) -> bool | tuple[bool, str]:
         Returns: False if sample.text does not pass any of the the heuristic tests
 
         """
-        from nltk.tokenize import word_tokenize
-
         text = doc.text
-        words = word_tokenize(text)  # TODO we should use language id filter
+        language = doc.metadata.get("language", "en")
```
vsabolcec, line 73: To keep backward compatibility, English tokenizer is used if no language metadata is provided. We could also use [multi-language](https://spacy.io/models/xx) tokenizer from Spacy or require that language metadata is present.

tests/pipeline/test_word_tokenizers.py
```diff
@@ -0,0 +1,29 @@
...
+            tokens = default_tokenizer.tokenize(SAMPLE_TEXT, language)
+            assert len(tokens) >= 1, f"'{language}' tokenizer assertion failed"
+            is_stripped = [token == token.strip() for token in tokens]
+            assert all(is_stripped), f"'{language}' tokens contain whitespaces"
+
+    def test_english_tokenizer(self):
+        nltk_words = word_tokenize(SAMPLE_TEXT, language="english")
+        tokenizer_words = default_tokenizer.tokenize(SAMPLE_TEXT, language=Languages.english)
+
+        self.assertEqual(nltk_words, tokenizer_words, "NLTK tokenizer and multilingual tokenizer differ")
```
vsabolcec, line 29: This test ensures that the "old" default tokenizer and "new" default tokenizer produce the same output. Is this test enough to ensure the correctness of the switch or do we also want to add integration tests?

vsabolcec:
@guipenedo the PR is ready for review

guipenedo (changes requested):
Great work! Some changes:
- move word_tokenizers.py to `utils` (`tools` is more for CLI tools)
- at some point we were planning to load the language from the metadata but have moved a bit to a more explicit approach: users should pass a `language` option to each block that uses tokenization of some sort. We believe this is less error prone than using the language in the metadata. The default value here should still be `en`
- given the above, not sure having the MultilingualTokenizer makes a lot of sense
- instead of a `tokenizer` option being passed to the blocks, users should instead pass the `language` option, and there could be a `load_tokenizer` function in word_tokenizers.py that would return the correct tokenizer (and possibly cache it on a dictionary or something so it can reused across blocks)
- there are many other blocks using sent_tokenize and word_tokenize but we can change those later

guipenedo (approved):

hynky1999 (commented):
LGTM!
Is there any need for the cache ?

hynky1999:
> LGTM! Is there any need for the cache ?

@guipenedo just explained. Approved

hynky1999 (approved):

PR #149, opened 2024-04-09
Add more languages

jordane95:


jordane95:
May be we could add all language codes supported by the fasttext classifier using third party lib like `langcodes`?

guipenedo:
Closed as we will merge #147 instead

PR #153, opened 2024-04-12
Adds `tasks_per_job` to slurm executor

guipenedo:
Adds an option to run multiple datatrove tasks in a single slurm job.

### DIFF
--- a/src/datatrove/executor/slurm.py
+++ b/src/datatrove/executor/slurm.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import math
 import os
 import random
 import signal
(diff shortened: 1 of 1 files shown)


src/datatrove/executor/slurm.py
```diff
@@ -79,7 +80,7 @@ class SlurmPipelineExecutor(PipelineExecutor):
         mail_type: see https://slurm.schedmd.com/sbatch.html. Common values are (NONE, BEGIN, END, FAIL, REQUEUE, ALL)
         mail_user: email address to send notifications to
         requeue: requeue the job if it fails
-
+        tasks_per_job: each slurm job in the job array will run this many datatrove tasks. This reduces the total nb of slurm jobs launched.
```
guipenedo, line 83:
```suggestion
        tasks_per_job: each slurm job in the job array will run these many datatrove tasks. This reduces the total nb of slurm jobs launched.
```

PR #154, opened 2024-04-15
Unsigned int tokenizer and srun args

marianna13:
Two suggestions:

1. Change unsigned short to to unsigned int in `struct.pack` [here](https://github.com/huggingface/datatrove/blob/209ebec293cb2407c8e6fc691bfeb358ba030b09/src/datatrove/pipeline/tokens/tokenizer.py#L153) because for me short didn't work with some tokenizers (e.g. [stablelm tokenizer](https://huggingface.co/stabilityai/stablelm-2-1_6b/tree/main)).
2. Add `srun_args` varibale to slurm executor. Sometimes I need to pass some additional args to srun due to specific cluster configurations (e.g. `kill-on-bad-exit`).

Thanks!


src/datatrove/pipeline/tokens/tokenizer.py
```diff
@@ -150,7 +150,7 @@ def write(self, tokens: list[int], loss_values: np.ndarray | None):
             loss_values (np.ndarray | None): optional loss values to write
         """
         # get the bytes for uint16 (H)
-        self.write_bytes(struct.pack("<%sH" % len(tokens), *tokens))
```
guipenedo, line 153: This has to be made optional (with the default being uint16 as before). There are also other changes required (size computations using "2" and the tokenizer merger, for example).

src/datatrove/executor/slurm.py
```diff
@@ -111,6 +111,7 @@ def __init__(
         mail_type: str = "ALL",
         mail_user: str = None,
         requeue: bool = True,
+        srun_args: dict = {},
```
guipenedo, line 114: Convention is usually to pass a sentinel value (None) as default instead of a mutable object

guipenedo (changes requested):

marianna13:
Thanks for comments! I did this trick with computing nu,ber of bytes and tokens type (so if maximum value of tokens is more then 2**16 so too big for uint16 we will use uint32 and 4 bytes per token consequently). I don't really like my implementation (I think there's a way to make it cleaner). But conceptually I guess this is what you meant. Please let me know your thoughts!

guipenedo:
Hi! sorry for the late reply. I actually forgot you were working on uint32 tokenization and ended up implementing it myself on #160 as we needed it for an internal project at HF, is it ok with you if we change the PR to only include the srun_args change? I can also add you as a co-author of #160 if you'd like :)

guipenedo (approved):

guipenedo:
I've kept only the srun args changes as the rest has already been added to tokenizers :)

PR #155, opened 2024-04-15
Fix race conditions during download/extraction

### LINKED ISSUE #140: Bug in url filter
```
2024-03-29 08:17:44.786 | INFO | datatrove.executor.local:_launch_run_for_rank:73 - 1/20 tasks completed.
2024-03-29 08:17:44.788 | INFO | datatrove.executor.local:_launch_run_for_rank:73 - 2/20 tasks completed.
multiprocess.pool.RemoteTraceback:
"""
Traceback (most recent call last):
File "/opt/conda/envs/datatrove/lib/python3.10/site-packages/multiprocess/pool.py", line 125, in worker
result = (True, func(*args, **kwds))
File "/output/datatrove/src/datatrove/executor/local.py", line 68, in _launch_run_for_rank
return self._run_for_rank(rank, local_rank)
File "/output/datatrove/src/datatrove/executor/base.py", line 96, in _run_for_rank
raise e
File "/output/datatrove/src/datatrove/executor/base.py", line 83, in _run_for_rank
deque(pipelined_data, maxlen=0)
File "/output/datatrove/src/datatrove/pipeline/writers/disk_base.py", line 173, in run
for document in data:
File "/output/datatrove/src/datatrove/pipeline/extractors/base.py", line 77, in run
for doc in data:
File "/output/datatrove/src/datatrove/pipeline/filters/base_filter.py", line 50, in run
filter_result, reason = get_filter_result(self.filter(doc))
File "/output/datatrove/src/datatrove/pipeline/filters/url_filter.py", line 87, in filter
self.download_data()
File "/output/datatrove/src/datatrove/pipeline/filters/url_filter.py", line 76, in download_data
tar.extractall(download_dir)
File "/opt/conda/envs/datatrove/lib/python3.10/tarfile.py", line 2264, in extractall
self._extract_one(tarinfo, path, set_attrs=not tarinfo.isdir(),
File "/opt/conda/envs/datatrove/lib/python3.10/tarfile.py", line 2331, in _extract_one
self._handle_fatal_error(e)
File "/opt/conda/envs/datatrove/lib/python3.10/tarfile.py", line 2327, in _extract_one
self._extract_member(tarinfo, os.path.join(path, tarinfo.name),
File "/opt/conda/envs/datatrove/lib/python3.10/tarfile.py", line 2402, in _extract_member
os.makedirs(upperdirs)
File "/opt/conda/envs/datatrove/lib/python3.10/os.py", line 225, in makedirs
mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/root/.cache/huggingface/assets/datatrove/filters/url_filter/adult'
"""
```

hynky1999:
This PR solves a race-condition as reported by https://github.com/huggingface/datatrove/issues/140, during extraction/downloading files.
The race-condition are resolved using inter-process file locks (`fasteners` lib).
The nice thing about `fasteners` is that they leverage syscalls, which should work across all the platforms (tested on mac/linux). This also means that even when the processes crashes all locks are guaranteed to be removed. 

Fixes #140

### DIFF
--- a/src/datatrove/io.py
+++ b/src/datatrove/io.py
@@ -2,6 +2,7 @@ import os.path
 from glob import has_magic
 from typing import IO, TypeAlias
 
+from fasteners import InterProcessLock
 from fsspec import AbstractFileSystem
 from fsspec import open as fsspec_open
 from fsspec.callbacks import NoOpCallback, TqdmCallback
@@ -300,6 +301,13 @@ def download_file(remote_path: str, local_path: str, progress: bool = True):
         if progress
         else NoOpCallback(),
     )
+
+
+def download_file_safely(remote_path: str, local_path: str, progress: bool = True):
+    with InterProcessLock(f"{local_path}.lock"):
+        if os.path.exists(local_path):
+            return
+        download_file(remote_path, local_path, progress)
 
 
 DataFolderLike: TypeAlias = str | tuple[str, dict] | DataFolder
--- a/src/datatrove/io.py
+++ b/src/datatrove/io.py
@@ -1,4 +1,3 @@
-import contextlib
 import os.path
 from glob import has_magic
 from typing import IO, Callable, TypeAlias
@@ -304,42 +303,29 @@ def download_file(remote_path: str, local_path: str, progress: bool = True):
     )
 
 
-@contextlib.contextmanager
-def safely_process(file: str, is_processed: Callable[[], bool]):
+def safely_create_file(file_to_lock: str, do_processing: Callable):
     """
-    Context manager for process-safe file manipulation.
-    Will check if the file has been processed and if not, will process-lock and let caller do the processing.
-    If the file has been processed, the lock will not be acquired and the context manager will be a no-op.
-    This can be verified by determining if the returned value is None.
+    Gets a lock to download/process and create some file(s). When processing is done a ".completed" file is created.
+    If this file already exists, we skip the processing. Otherwise, we try to acquire a lock and when we get it if the
+    completed file has not been created yet, we run the processing.
+
     Args:
-        file: str: The file to process
-        is_processed: Callable[[], bool]: A function that checks if the file has been processed
-    Yields:
-        None if no processing is needed (file is already processed)
-        Lock object if processing is needed
+        file_to_lock: str: lock will be "lock_path.lock" and completed file "lock_path.completed"
+        do_processing: callback with the code to run to process/create the files
     """
     from fasteners import InterProcessLock
 
-    lock = InterProcessLock(f"{file}.lock")
+    completed_file = f"{file_to_lock}.completed"
 
-    # If the processing was already done and no lock exists, we exit prematurely.
-
-    # AFAIK, there is no way to check if there is existing lock in win32 api,
-    # while this is possible on POSIX Linux, fcntl doesn't support such syscall for some reason,
-    # https://github.com/python/cpython/issues/96694.
-    # We thus check for file presence. This can have false positives as the file might not be cleaned properly.
-    if not lock.exists() and is_processed():
-        yield None
+    # if the completed file exists, we exit straight away
+    if os.path.exists(completed_file):
         return
 
-    with lock:
-        # Make sure to do  check process-locked as otherwise race condition can happen, and processes,
-        # which don't perform the download, could try to open the file before it's fully downloaded
-        if is_processed():
-            yield None
-            return
-
-        yield lock
+    # file is either being downloaded or needs to be downloaded
+    with InterProcessLock(f"{file_to_lock}.lock"):
+        if not os.path.exists(completed_file):
+            do_processing()
+            open(completed_file, "a").close()
 
 
 def cached_asset_path_or_download(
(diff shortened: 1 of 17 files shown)


src/datatrove/io.py
```diff
@@ -302,4 +303,21 @@ def download_file(remote_path: str, local_path: str, progress: bool = True):
...
+    This function is process-safe and will only download the file if it hasn't been downloaded already.
+    Args:
+        remote_path: str: The remote path to the file to download
+        local_path: str: The local path to save the file to
+        progress: bool: Whether to show a progress bar (Default value = True)
+    """
+    with InterProcessLock(f"{local_path}.lock"):
+        # Make sure to do exists check process-locked as otherwise race condition can happen, and processes,
+        # which don't perform the download, could try to open the file before it's fully downloaded
+        if os.path.exists(local_path):
```
guipenedo, line 318: How much does this slow down jobs when the file has already been downloaded previously and we have many tasks? Is there maybe some other smart combination of checks we could do (like check if the file to download exists AND the lock file does not exist) before creating a lock?

src/datatrove/pipeline/filters/fasttext_filter.py
```diff
@@ -77,7 +77,7 @@ def model(self):
             model_file = os.path.join(download_dir, strip_protocol(self.model_url).replace("/", "_"))
             if not os.path.isfile(model_file):
```
guipenedo, line 78: this is not consistent with the comment within the lock, i.e. running the check after acquiring the lock. I suppose ideally there should be a single line here which would be a call to a function that would do everything from checking if the file exists, to downloading it with the lock and printing the messages

src/datatrove/pipeline/filters/language_filter.py
```diff
@@ -48,7 +48,7 @@ def model(self):
             model_file = os.path.join(download_dir, "lid.176.bin")
             if not os.path.isfile(model_file):
```
guipenedo, line 49: same as above

src/datatrove/pipeline/filters/url_filter.py
```diff
@@ -69,11 +70,16 @@ def download_data(self):
         if self._downloaded:
             return
         download_dir = cached_assets_path(library_name="datatrove", namespace="filters", subfolder="url_filter")
-        if not os.path.isfile(os.path.join(download_dir, "adult", "domains")) or not os.path.isfile(
-            os.path.join(download_dir, "adult", "urls")
+
+        with (
+            InterProcessLock(os.path.join(download_dir, "url_filterblacklists.tar.gz.lock")),
+            tarfile.open(os.path.join(ASSETS_PATH, "url_filterblacklists.tar.gz"), "r:gz") as tar,
```
guipenedo, line 76: do we need to open the tarfile before checking if the files already exist?

src/datatrove/pipeline/filters/fasttext_filter.py
```diff
@@ -40,7 +40,7 @@ class FastTextClassifierFilter(BaseFilter):
     """
 
     name = "🤖 fastText"
-    _requires_dependencies = [("fasttext", "fasttext-wheel")]
+    _requires_dependencies = [("fasttext", "fasttext-wheel", "fasteners")]
```
guipenedo, line 43:
```suggestion
    _requires_dependencies = [("fasttext", "fasttext-wheel"), "fasteners"]
```

src/datatrove/pipeline/filters/language_filter.py
```diff
@@ -15,7 +15,7 @@
 
 class LanguageFilter(BaseFilter):
     name = "🌍 Language ID"
-    _requires_dependencies = [("fasttext", "fasttext-wheel")]
+    _requires_dependencies = [("fasttext", "fasttext-wheel", "fasteners")]
```
guipenedo, line 18:
```suggestion
    _requires_dependencies = [("fasttext", "fasttext-wheel"), "fasteners"]
```

guipenedo (changes requested):

src/datatrove/pipeline/filters/fasttext_filter.py
guipenedo: this fixes the tests

src/datatrove/pipeline/filters/fasttext_filter.py
hynky1999: mb

src/datatrove/pipeline/filters/url_filter.py
hynky1999: valid point

src/datatrove/io.py
hynky1999: Tested the following pipeline on 1k tasks:
```
        [
                lambda x,y,z: [Document("","", metadata={"url": "a.cz"})],
                URLFilter(),
        ],
```

No interprocess-guard:
```
real    1m35.601s
user    0m11.359s
sys     0m5.849s
```

With interprocess-guard:
```
real    2m29.377s
user    0m11.822s
sys     0m3.853s
```

So I implemented your idea. Bear in mind that it's not 100% bulettproof as I check for existence of lock using simple fs check instead of syscall (cause win don't support it and fnctl support in python is scuffed)

guipenedo: thought of a potentially safer approach: instead of checking if `file` exists, we can check if `file.processed` (or `file.downloaded`, something like this) exists, which is an empty file created only after the download is completed.

hynky1999: The current approach will work, there will be no race condition/ deadlock whatsoever. The only problem is that if the lock is not deleted, all processes will have to sequentially in this section instead of skipping it completely and once they do and remove the lock from fs, the next run will be able to skip the section completely. So this bad situation happens only if we crash and only the first run after the crash.

guipenedo (approved):

PR #158, opened 2024-04-20
Summary stats

hynky1999:
### This PR adds blocks for computing summary statistics
- Added base block for creating the summary statistics
- Added merging block for merging summary statistics computed in distributed manner.
- Blocks for: line, document, word, token, contamination level stats.
- Tests checking that both base block + merger works also tests for individual summary stats blocks.
- Example how to use
- Readme update

#### How stats works
- Each stat block has to implement `extract_stats`, which defines what statistics to track
- Statistics are then grouped based on:
* summary -> All documents together
* fqdn -> documents grouped by fqdn
* suffix -> documents grouped by suffix
* histogram -> grouping is done based on values of statistics

### Misc
- fixes bug in `from_dict` at `src/datatrove/utils/stats.py`


src/datatrove/utils/stats.py
```diff
@@ -304,7 +308,9 @@ def from_dict(cls, data):
         if isinstance(data, dict):
             total = data.get("total")
             mean = data.get("mean", 1)
-            n = data.get("n", total if mean != 1 else 1)
+            # We save n if we it has been added 1+ times and we didn't add just 1 -> mean == 1
+            # This means that if mean == 1 and we don't have n, the n must be total, otherwise 1
+            n = data.get("n", total if mean == 1 else 1)
```
guipenedo, line 313: good catch

src/datatrove/pipeline/stats/summary_stats/contamination_stats.py
```diff
@@ -0,0 +1,45 @@
...
+        super().__init__(output_folder, groups_to_compute, histogram_round_digits, top_k_config=top_k_config)
+        if len(words) == 0:
+            raise ValueError("At least one word must be provided")
+
+        self.words = words
+
+    def extract_stats(self, doc: Document) -> dict[str, int | float]:
+        from nltk.tokenize import word_tokenize
+
+        doc_words = word_tokenize(doc.text)
```
guipenedo, line 41: maybe add an option to normalize the text/use `simplify_test`

src/datatrove/pipeline/stats/summary_stats/lang_stats.py
```diff
@@ -0,0 +1,34 @@
...
+        histogram_round_digits: int = 3,
+        groups_to_compute: list[GROUP] = list(get_args(GROUP)),
+        top_k_config: TopKConfig = DEFAULT_TOP_K_CONFIG,
+    ) -> None:
+        super().__init__(output_folder, groups_to_compute, histogram_round_digits, top_k_config)
+        self.fasttext = FastTextModel([language])
+        self.language = language
+
+    def extract_stats(self, doc: Document) -> dict[str, int | float]:
+        return {f"fasttext_{self.language}": self.fasttext.predict(doc)[1][self.language]}
```
guipenedo, line 34: maybe optionally load from metadata instead if the value is there

src/datatrove/pipeline/stats/summary_stats/doc_stats.py
```diff
@@ -0,0 +1,38 @@
...
+    Summary stats of document level metrics:
+
+    Available stats:
+    length: Length of the document
+    white_space_ratio: Ratio of whitespace characters
+    non_alpha_digit_ratio: Ratio of non-alphabetic and non-digit characters
+    digit_ratio: Ratio of digits
+    """
+
+    type = "📊 - STATS"
```
guipenedo, line 20: this should only be on the base class

tests/pipeline/test_stats.py
```diff
@@ -0,0 +1,253 @@
...
+        top_k_config = TopKConfig(top_k=1, top_k_groups=["fqdn"])
+        summary_stats = DummyStats(output_folder=self.tmp_dir, top_k_config=top_k_config)
+        list(summary_stats.run(DOCS, 0, 1))
+
+        with self.tmp_dir.open("fqdn/stat/00000.json") as f:
+            stats = MetricStatsDict.from_dict(json.load(f))
+            self.assertEqual(stats["test1.co.uk"].total, 3.5)
+            self.assertEqual(stats["test2.cz"].total, 0)
+
+    def test_meging_stats(self):
```
guipenedo, line 101:
```suggestion
    def test_merging_stats(self):
```

guipenedo (commented):
Pretty nice, maybe there's no need to keep the previous stats/ module and we could just have this be there instead of inside stats/summary_stats?

src/datatrove/pipeline/stats/summary_stats/lang_stats.py
hynky1999: I don't like that, because typical way that language_score gets into metadata is through LID filter. However the filter will only save the most likely langauge/score. This would than yield a lot of 0, just because the langauge wasn't the most probable.

guipenedo: Why not check if the language in the metadata is the one you want? If it isn't then you can compute

src/datatrove/pipeline/stats/__init__.py
```diff
@@ -1,2 +1,11 @@
...
+from datatrove.pipeline.stats.contamination_stats import WordsContaminationStats
+from datatrove.pipeline.stats.doc_stats import DocStats
+from datatrove.pipeline.stats.lang_stats import LangStats
+from datatrove.pipeline.stats.line_stats import LineStats
+from datatrove.pipeline.stats.merger import STATS_MERGED_NAME, StatsMerger
+from datatrove.pipeline.stats.paragraph_stats import ParagraphStats
+from datatrove.pipeline.stats.perplexity_stats import CCNetPerplexityStats
+from datatrove.pipeline.stats.sentence_stats import SentenceStats
+from datatrove.pipeline.stats.token_stats import TokenStats
+from datatrove.pipeline.stats.word_stats import WordStats
```
guipenedo, line 11: won't the absolute imports here give an error about partially initialized modules?

src/datatrove/pipeline/stats/base.py
```diff
@@ -0,0 +1,112 @@
...
+        raise NotImplementedError()
+
+    def get_kv(self, doc: Document, value: STAT_TYPE, group_name: GROUP) -> tuple[str, STAT_TYPE]:
+        if group_name == "histogram":
+            # Use rounding to reduce then number of values for histogram
+            return str(round(value, self.histogram_round_digits)), 1
+        elif group_name == "summary":
+            return "summary", value
+        elif group_name == "fqdn":
+            return doc.metadata.setdefault("fqdn", self.tld_extractor.extract_str(doc.metadata["url"]).fqdn), value
```
guipenedo, line 69: does this not rerun tld_extractor every time, including when fqdn is already there?

src/datatrove/pipeline/stats/base.py
```diff
@@ -0,0 +1,112 @@
...
+                    logger.error(f"Error while extracting stats from document {doc.id}", exc_info=e)
+                    raise e
+
+                for group, counters in groups_dicts.items():
+                    for stat, value in doc_stats.items():
+                        key, value = self.get_kv(doc, value, group)
+                        counters[stat][key] += value
+
+                doc.metadata.update(doc_stats)
+                yield doc
```
guipenedo, line 93: this should be outside the block with track time (put the track time inside the loop)

src/datatrove/pipeline/stats/__init__.py
hynky1999: No they won't as they don't depend on each other

src/datatrove/pipeline/stats/base.py
guipenedo: It's still inside the track_time xD
```suggestion
            yield doc
```

Issue #168, opened 2024-05-03
Custom Adapter Function Cannot Use self in BaseReader Class


justHungryMan:
### Problem
Currently, the `BaseReader` class accepts a `adapter` parameter that is used to transform input data into the required document format. However, if a custom adapter function is used, it does not have access to the `BaseReader` instance variables, limiting its flexibility and requiring workarounds.

### Suggested Enhancement
I propose modifying the assignment of the `adapter` to use `MethodType` from the `types` module. This change will allow the custom adapter to access instance variables and methods, greatly increasing its utility and aligning with common Python practices for class method customization.

### Example Use Case
Consider a scenario where the adapter needs to access `text_key` and `id_key` from the reader instance to properly format the input data. With the proposed change, this becomes straightforward and the code remains clean and maintainable.

This change would make the `BaseReader` more versatile and user-friendly for complex data loading scenarios.

I would appreciate feedback on the proposed change.

PR #169, opened 2024-05-03
Enhance BaseReader to allow custom adapters access to instance variables

justHungryMan:
This PR addresses the issue #168 by modifying how the `adapter` is assigned within the `BaseReader` class. By using `MethodType`, we enable custom adapters to access instance variables and methods, enhancing the flexibility and usability of custom adapters.

### Changes Made
- Changed the assignment of `self.adapter` to `MethodType(adapter, self)` if a custom adapter is provided.
- Updated relevant unit tests to reflect this change and ensure functionality.

### Impact
This change allows developers to write more intuitive and robust custom adapters by leveraging the internal state of the `BaseReader` instance, leading to cleaner and more maintainable code.

### Tests
- Added a test case in `tests/pipeline/test_adapter_reader.py` to verify that the custom adapter can access and modify instance variables as expected.

Please review the changes and provide your feedback.

### DIFF
--- a/src/datatrove/pipeline/readers/base.py
+++ b/src/datatrove/pipeline/readers/base.py
@@ -1,6 +1,7 @@
 import random
 from abc import abstractmethod
 from contextlib import nullcontext
+from types import MethodType
 from typing import Callable
 
 from loguru import logger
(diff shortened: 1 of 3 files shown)


guipenedo:
Maybe we should also add this to writers' adapters?

justHungryMan:
Do we need test code for writer's adapter?

guipenedo:
No I don't think so

guipenedo:
I added it and also changed the docstrings a bit.
Thanks a lot, great change :)

PR #170, opened 2024-05-03
Hf dataset update

hynky1999:
### Updates the Huggingface dataset to improve performance and fix bugs
- It's now possible to use dataset streaming
- Adds an error message if the dataset is of type `DatasetDict` instead of raising general AttributeError
- Fixes the bug which occurred when no `dataset options were specified`


src/datatrove/pipeline/readers/huggingface.py
```diff
@@ -44,23 +47,53 @@ def __init__(
...
+                info=dst._info.copy(),
+                split=dst._split,
+                formatting=dst._formatting,
+                shuffling=copy.deepcopy(dst._shuffling),
+                distributed=copy.deepcopy(dst._distributed),
+                token_per_repo_id=dst._token_per_repo_id,
+            )
+        else:
+            # If we have just a single shard/file, we shard inter-file
+            return split_dataset_by_node(dst, rank, world_size)
```
lhoestq, line 81: how is this different that just calling `split_dataset_by_node` ?

hynky1999: Yes exactly, we have this same behavior for other readers so I wanted to be consistent.

PR #174, opened 2024-05-05
Optimize URLFilter and add option to disable integrated wordlists

its5Q:
This pull request makes two changes to URLFilter:
1. Replacing the naive subword search with a faster Aho-Corasick algorithm for multiple subwords.
2. Adding an argument that allows users to disable the usage of integrated block lists so they can supply their own lists instead.

### DIFF
--- a/src/datatrove/pipeline/filters/url_filter.py
+++ b/src/datatrove/pipeline/filters/url_filter.py
@@ -56,8 +56,8 @@ class URLFilter(BaseFilter):
         use_integrated_lists: bool = True,
         exclusion_writer: DiskWriter = None,
     ):
-        from tldextract import TLDExtract
         import ahocorasick
+        from tldextract import TLDExtract
 
         super().__init__(exclusion_writer)
         self.soft_word_threshold = soft_word_threshold
--- a/src/datatrove/pipeline/filters/url_filter.py
+++ b/src/datatrove/pipeline/filters/url_filter.py
@@ -27,7 +27,7 @@ def parse_list(line, do_normalize=True):
 
 def get_list(abs_path: str, file_name: str, extra: set = None, do_normalize: bool = True):
     with open(os.path.join(abs_path, file_name)) as f:
-        return parse_list(f, do_normalize).union(set(parse_list(extra, do_normalize)) if extra else set())
+        return parse_list(f, do_normalize).union(extra)
 
 
 class URLFilter(BaseFilter):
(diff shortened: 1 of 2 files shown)


guipenedo (changes requested):
Great changes, indeed we should have added a way to disable the default lists before. The perf improvement with ahocorasick is also nice

src/datatrove/pipeline/filters/url_filter.py
```diff
@@ -53,22 +53,32 @@ def __init__(
...
         from tldextract import TLDExtract
 
         super().__init__(exclusion_writer)
         self.soft_word_threshold = soft_word_threshold
-        self.block_listed_domains = extra_domains
-        self.block_listed_url = extra_urls
-        self.banned_words = banned_words
-        self.banned_subwords = banned_subwords
-        self.soft_banned_words = soft_banned_words
+        self.block_listed_domains = parse_list(extra_domains, do_normalize=False) if extra_domains else set()
```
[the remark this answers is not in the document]
its5Q: Yeah, I think it's now safe to remove as it's guaranteed to be a set.

its5Q:
Also added a check if the block list is empty, otherwise it would fail with empty block list because the aho-corasick tree isn't built.

src/datatrove/pipeline/filters/url_filter.py
```diff
@@ -27,7 +27,7 @@ def parse_list(line, do_normalize=True):
 
 def get_list(abs_path: str, file_name: str, extra: set = None, do_normalize: bool = True):
```
guipenedo, line 28:
```suggestion
def get_list(abs_path: str, file_name: str, extra: set, do_normalize: bool = True):
```

src/datatrove/pipeline/filters/url_filter.py
```diff
@@ -53,22 +53,32 @@ def __init__(
...
+        self.block_listed_url = parse_list(extra_urls, do_normalize=False) if extra_urls else set()
+        self.banned_words = parse_list(banned_words) if banned_words else set()
+        self.banned_subwords = parse_list(banned_subwords) if banned_subwords else set()
+        self.soft_banned_words = parse_list(soft_banned_words) if soft_banned_words else set()
+        self.use_integrated_lists = use_integrated_lists
         self._downloaded = False
         self.tldextractor = TLDExtract()
 
+        self.banned_subwords_automaton = ahocorasick.Automaton(ahocorasick.STORE_INTS)
+        for word in self.banned_subwords:
```
guipenedo, line 74: this probably isn't an issue but if `use_integrated_lists=True`, won't the extra `banned_subwords` be added twice? Once here and again inside `download_data`

its5Q: that's probably fine, if the trie already exists add_word should just return False

guipenedo (approved):

guipenedo:
Ty :)

Issue #163, opened 2024-04-28
Log progress


jordane95:
Can we add a logging function to print out how much files we have processed? So that we could estimate the total runtime of one job dynamically

PR #176, opened 2024-05-07
Add progres for files

hynky1999:
Adds progress bar also to for file as in https://github.com/huggingface/datatrove/issues/163.
Additionaly adds better description to progress bars

fixes https://github.com/huggingface/datatrove/issues/163

### DIFF
--- a/src/datatrove/pipeline/readers/base.py
+++ b/src/datatrove/pipeline/readers/base.py
@@ -1,6 +1,5 @@
 import random
 from abc import abstractmethod
-from contextlib import nullcontext
 from typing import Callable
 
 from loguru import logger
@@ -17,7 +16,8 @@ class BaseReader(PipelineStep):
 
     Args:
         limit: limit the number of documents to read. Useful for debugging
-        progress: show tqdm progress bar. Might be spammy in some environments
+        file_progress: show tqdm progress bar for files. Might be spammy in some environments
+        doc_progress: show tqdm progress bar for documents. Might be spammy in some environments
         adapter: function to adapt the data dict from the source to a Document.
             Take as input: data: dict, path: str, id_in_file: int | str
             Return: a dict with at least a "text" key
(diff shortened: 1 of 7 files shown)


guipenedo:
Might be a bit too spammy, maybe a `doc_progress` bool and a `file_progress` one?

src/datatrove/pipeline/readers/base.py
```diff
@@ -185,10 +190,19 @@ def read_files_shard(self, shard: list[str]) -> DocumentsPipeline:
...
+                unit="doc",
+                disable=not self.doc_progress,
+            ) as doc_pbar,
+            tqdm(total=len(shard), desc="File progress", unit="file", disable=not self.file_progress) as file_pbar,
+        ):
+            for i, filepath in enumerate(shard):
                 self.stat_update("input_files")
-                logger.info(f"Reading input file {filepath}")
+                logger.info(f"Reading input file {filepath}, {i+1}/{len(shard)}")
+                file_pbar.update()
```
guipenedo, line 205: I think this should be after the loop with all the docs, so that we only show one file processed after we finish the entire file

src/datatrove/pipeline/readers/huggingface.py
```diff
@@ -36,13 +35,13 @@ def __init__(
...
         batch_size: int = 1000,
-        progress: bool = False,
+        doc_progress: bool = False,
         adapter: Callable = None,
         text_key: str = "text",
         id_key: str = "id",
         default_metadata: dict = None,
     ):
-        super().__init__(limit, skip, progress, adapter, text_key, id_key, default_metadata)
+        super().__init__(limit, skip, False, doc_progress, adapter, text_key, id_key, default_metadata)
```
guipenedo, line 44: maybe we should move the progress arguments from `BaseReader` to `BaseDiskReader` so that we don't have to pass False here which doesn't make much sense

src/datatrove/pipeline/readers/base.py
hynky1999: Not sure if I understand. You think it would be better to update the progress bar after we actually processed the file ?

guipenedo: exactly

guipenedo (approved):

PR #179, opened 2024-05-07
Migrate dedup to xxhash

guipenedo:
- Introduces new class `HashConfig`, used for specifying how what hashfc/precission to use
- Migrates all classes to the new config for consistency
- Fixing bug in `SentenceDedup`. The bug occurs when hashes from first stage don't cover the whole hash space. This results in non-correspondance between file_name and it's order (id) in signatures folder. However in second order we rely on this correspondance as we assume `id.sig == file_name`. I fixed this by relying solely on the file_name. The same bug occurs in `url_dedup`. Lastly I made the same change to `minhash`, where this bug shouldn't occur, but it makes more sense to not assume this correspondance.
- Separated hashing code to new files
- To make `xxhash` optional, we no longer can use `requires....` on class blocks as it's possible to also use `sha1` there. The check for dependency is thus done when the new hash function is created. See: `create_hash_func`
- Added a decorator for testing the hash configs. By default it only runs with 32/64b precission and don't alter the hash function, the reasoning is to save a bit of time as we don't assume that the hash_fc change should break anything.

### DIFF
--- a/src/datatrove/pipeline/dedup/minhash.py
+++ b/src/datatrove/pipeline/dedup/minhash.py
@@ -453,7 +453,9 @@ class MinhashDedupCluster(PipelineStep):
         def parent(x):
             if x not in union_set or union_set[x] == x:
                 return x
-            return parent(union_set[x])
+            # Path Compression
+            union_set[x] = parent(union_set[x])
+            return union_set[x]
 
         with self.track_time():
             for dup_file in dup_files:
--- a/src/datatrove/pipeline/dedup/minhash.py
+++ b/src/datatrove/pipeline/dedup/minhash.py
@@ -74,7 +74,7 @@ class HashSig:
 
     sig: tuple[int]
     file_id: int
-    file_name: int
+    file_stem: str
     doc_id: int
     reader_id: int
 
@@ -106,7 +106,7 @@ def read_sigs(
             return
         seek_to_start(f, min_hash, line_format, config.hash_config.struct_format)
         last = None
-        file_name = int(Path(file.path).name.replace(".minhash.sig", ""))
+        file_stem = Path(file.path).name.replace(".minhash.sig", "")
         for data in read_tuples_from_file(f, line_format, lines_to_buffer=lines_to_buffer):
             sigdata = data if index_file else data[:-1]
             assert sigdata[0] >= min_hash and (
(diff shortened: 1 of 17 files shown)


guipenedo:
> To make xxhash optional, we no longer can use requires.... on class blocks as it's possible to also use sha1 there. The check for dependency is thus done when the new hash function is created. See: create_hash_func

The new thing is fine but we could just add it to the list of dependencies based on HashConfig from the __init__

src/datatrove/pipeline/dedup/minhash.py
```diff
@@ -111,12 +96,13 @@ def read_sigs(
...
     """
-    line_format = f"{config.hashes_per_bucket}{config.hash_format}{'I' if not index_file else ''}"
+    line_format = f"{config.hashes_per_bucket}{config.hash_config.struct_format}{'I' if not index_file else ''}"
     with file as f:
         if f.size == 0:
             return
-        seek_to_start(f, min_hash, line_format, config.hash_format)
+        seek_to_start(f, min_hash, line_format, config.hash_config.struct_format)
         last = None
+        file_stem = Path(file.path).name.replace(".minhash.sig", "")
```
guipenedo, line 105:
```suggestion
        file_stem = Path(file.path).name.removesuffix(".minhash.sig")
```

src/datatrove/pipeline/dedup/sentence_dedup.py
```diff
@@ -157,18 +163,23 @@ def run(self, data: DocumentsPipeline, rank: int = 0, world_size: int = 1):
...
-    file: AbstractBufferedFile, file_id: int, index_file: bool = False, lines_to_buffer: int = 5
+    file: AbstractBufferedFile,
+    file_id: int,
+    config: SentDedupConfig,
+    index_file: bool = False,
+    lines_to_buffer: int = 5,
 ) -> Generator[HashSig, None, None]:
-    line_format = "QIH" if not index_file else "Q"
+    line_format = f"{config.hash_config.struct_format}IH" if not index_file else config.hash_config.struct_format
+    file_stem = Path(file.path).name.replace(ExtensionHelperSD.stage_1_signature, "")
```
guipenedo, line 173:
```suggestion
    file_stem = Path(file.path).name.removesuffix(ExtensionHelperSD.stage_1_signature)
```

src/datatrove/pipeline/dedup/url_dedup.py
```diff
@@ -154,20 +155,23 @@ def run(self, data: DocumentsPipeline, rank: int = 0, world_size: int = 1):
...
     file_id: int,
+    hash_config: HashConfig,
     index_file: bool = False,
     lines_to_buffer: int = 5,
 ) -> Generator[HashSig, None, None]:
     last = None
-    line_format = "QHI" if not index_file else "Q"
+    line_format = f"{hash_config.struct_format}HI" if not index_file else hash_config.struct_format
     with file as f:
+        file_stem = Path(f.path).name.replace(ExtensionHelperSD.stage_1_signature, "")
```
guipenedo, line 165:
```suggestion
        file_stem = Path(f.path).name.removesuffix(ExtensionHelperSD.stage_1_signature)
```

PR #181, opened 2024-05-10
remove ListFilter from the process_common_crawl_dump example

QasidSaleem:
Removes ListFilter from the examples/process_common_crawl_dump.py as it is no longer available in the datatrove.pipeline.filters.


hynky1999:
Good sport, we have moved to to fineweb filters!
Thanks.

Issue #180, opened 2024-05-08
how to turn log/traceback color off?


stas00:
Trying datatrove for the first time and the program spews a bunch of logs and tracebacks in yellow and cyan which are completely unreadable on the b&w console. 

Does the program make an assumption that the user is using w&b (dark) console?

I tried to grep for `color` to see how it controls the colors but found nothing relevant, so it's probably some 3rd party component that does that.

If the coloring logic doesn't bother to check what the console colors are to keep the output readable, any idea how to turn it off completely? I RTFM'ed - didn't find any docs that address that aspect.

Thanks a lot!

guipenedo:
Hi,
Indeed that's an oversight on our part, we'll add an option somewhere. 
For now I think if you set the env variable `export LOGURU_COLORIZE=NO` it should disable them
Sorry about that and thank you for taking an interest in datatrove :)

guipenedo:
For reference it's this line here with `colorize=True`:
https://github.com/huggingface/datatrove/blob/main/src%2Fdatatrove%2Futils%2Flogging.py#L46

stas00:
I don't think it's `colorize=True`, it's the first thing I tried to set to `False` and it made no difference. I have retested it now, the color is still there. 

But `LOGURU_COLORIZE=NO` did the trick, thank you very much, @guipenedo!

It looks like even the package that caused the problem isn't documenting this feature
https://github.com/search?q=repo%3ADelgan%2Floguru%20LOGURU_COLORIZE&type=code

Would it be kindly possible to document this somewhere in this project's docs? Thank you!

stas00:
I confirm that it works.

May I suggest that this is a sub-optimal way to implement this - as now the user has to pass this flag multiple times and the main issue is that this flag has nothing to do with the Executor but the library itself. This is a global behavior - either the user's setup can support the rainbow colors or it doesn't. It's a one-time binary decision.

Typically HF libraries have `logging.py` where logging levels and behavior is control'ed this would allow the user to set the logging behavior once at the beginning of the program and be done with it. e.g. see https://github.com/huggingface/transformers/blob/main/src/transformers/utils/logging.py
This approach works really well and ideally should belong there.

Thank you.

guipenedo:
Very well, point taken.
I've removed the options from the executor and added the following env variables:
- `DATATROVE_COLORIZE_LOGS` "1" to add ANSI colors to console log messages and "0" to disable colorization.
[...]

PR #184, opened 2024-05-13
document how to turn off colorization

stas00:
Fixes: https://github.com/huggingface/datatrove/issues/180

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,11 @@ For a pipeline with `logging_dir` **mylogspath/exp1**, the following folder stru
 ```
 </details>
 
+### Colors
+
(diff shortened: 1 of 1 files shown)


guipenedo:
I'd rather fix this in some other way without requiring changes from the user. Can you share some more details about the terminal you are using? Also, is the issue with the logs printed directly to the console or also when opening/reading the saved log files?

stas00:
KDE konsole Black on White - but I'm pretty sure it'd impact any light mode setup.

snapshot_763

I'm not an expert on setting colors - but getting the fore/background colors right automatically is a non-trivial task... probably doing non-colorized output out of the box and offering the colorization as an optional enhancement where the user provides their b&w or w&b mode as an argument might be a low hanging fruit if there could be a color map used for each of the major modes.

guipenedo:
Closed as issue was addressed on #185

PR #189, opened 2024-05-22
Migrate pipeline blocks to new word tokenizers

guipenedo:
- adds dependencies checking to word tokenizers
- lazy load tokenizers
- change all `word_tokenize`, `sent_tokenize` etc in the codebase to use the new multilingual word tokenizers


hynky1999 (approved):
LGTM!
One problem that I see with our new word tokenizers, is that we use don't use process locks for downloading the models (if they are needed), this can introduce problems when multiple processes start at the same time as we have seen before.

However I don't think that's something to do now. I have created an issue https://github.com/huggingface/datatrove/issues/190

PR #193, opened 2024-05-24
Extend randomize_start feature to local executor

justHungryMan:
Extend the randomize_start option, which is available in the slurm executor, to the local executor. Initially, it may seem that randomize_start is unnecessary for local operations. However, specific scenarios, such as processing WARCs directly from Common Crawl's S3, highlight its importance. When dealing with S3, there is a risk of exceeding the permitted rate limit for API calls resulting in potential blocking, if to much API calls are occured simultaneously. By enabling either sequential or randomized execution through the executor, we can mitigate this issue.

The addition of this feature to the local executor was prompted by running the fineweb examples locally.

### DIFF
--- a/src/datatrove/executor/base.py
+++ b/src/datatrove/executor/base.py
@@ -1,5 +1,7 @@
 import dataclasses
 import json
+import random
+import time
 from abc import ABC, abstractmethod
 from collections import deque
 from collections.abc import Sequence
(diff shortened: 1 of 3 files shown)


guipenedo:
Maybe we could move the sleep inside the base class in this case, since both local and slurm executors will be using it?

justHungryMan:
@guipenedo That could be also a great choice

PR #201, opened 2024-05-29
Add an example for filtering an HF dataset and push to hub

loubnabnl:


guipenedo:
The linting needs fixing. Can you also add a comment at the top of the code roughly explaining how this works?

guipenedo:
Merged :)

PR #203, opened 2024-05-30
Issues w/ DatatroveFolderDataset

TJ-Solergibert:
Hi!

While using `DatatroveFolderDataset` I encountered the following:
1. While using a `LocalFileSystem`, a folder containing tokenised documents and using a `filename_pattern` the call to [`fs.glob(...)`](https://github.com/huggingface/datatrove/blob/0f2c69f8249aa0c53ebcf10afa2394da506a953f/src/datatrove/utils/dataset.py#L101) fails because it's expecting a object with [`.endswith()`](https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/spec.html#AbstractFileSystem.glob) method not a `Pattern`. Here is an example:

```python
import re
from fsspec.core import url_to_fs

fs, folder_path = url_to_fs("src/datatrove")

pattern = "src/datatrove/*.py"
filename_pattern = re.compile(pattern)

fs.glob(filename_pattern, maxdepth=1)

AttributeError                            Traceback (most recent call last)
Cell In[31], [line 1](vscode-notebook-cell:?execution_count=31&line=1)
----> [1](vscode-notebook-cell:?execution_count=31&line=1) fs.glob(filename_pattern, maxdepth=1)

File ~/.local/lib/python3.10/site-packages/fsspec/spec.py:564, in AbstractFileSystem.glob(self, path, maxdepth, **kwargs)
     import re
     seps = (os.path.sep, os.path.altsep) if os.path.altsep else (os.path.sep,)
     ends_with_sep = path.endswith(seps)  # _strip_protocol strips trailing slash
     path = self._strip_protocol(path)
     append_slash_to_dirname = ends_with_sep or path.endswith(tuple(sep + "**" for sep in seps))

AttributeError: 're.Pattern' object has no attribute 'endswith'

```
  Using the `str` of the `Pattern` object solves the issue, but then looks a bit useless the [`re.compile`](https://github.com/huggingface/datatrove/blob/0f2c69f8249aa0c53ebcf10afa2394da506a953f/src/datatrove/utils/dataset.py#L95) right?
```
fs.glob(filename_pattern.pattern, maxdepth=1)

['/mloscratch/homes/solergib/nanotrove/datatrove/src/datatrove/__init__.py',
 '/mloscratch/homes/solergib/nanotrove/datatrove/src/datatrove/data.py',
 '/mloscratch/homes/solergib/nanotrove/datatrove/src/datatrove/io.py']
```

I'm using `re==2.2.1` & `fsspec==2024.3.1`!

2. Should we take care of closing the [opened files](https://github.com/huggingface/datatrove/blob/0f2c69f8249aa0c53ebcf10afa2394da506a953f/src/datatrove/utils/dataset.py#L51)?
3. It's cool to loop back to the beginning of the dataset if we index a position greater than `len(dataset)` but I can't see when this will happen if both `__len__` methods of `DatatroveFolderDataset` & `DatatroveFileDataset` return the real length (Used by the sampler for example). Maybe we are hiding indexing positions out of bounds, right?

### DIFF
--- a/src/datatrove/utils/dataset.py
+++ b/src/datatrove/utils/dataset.py
@@ -1,7 +1,4 @@
-import re
 from bisect import bisect
-from re import Pattern
-from typing import Union
 
 import numpy as np
 import torch
@@ -83,7 +80,7 @@ if is_torch_available():
             self,
             folder_path: str,
             seq_len: int,
-            filename_pattern: Union[Pattern, str] = None,
+            filename_pattern: str = None,
             recursive: bool = True,
             token_size: int = 2,
             max_tokens: int | None = None,
@@ -91,14 +88,12 @@ if is_torch_available():
             seed: int = 42,
         ):
             self.folder_path = folder_path
-            if isinstance(filename_pattern, str):
-                filename_pattern = re.compile(filename_pattern)
             self.filename_pattern = filename_pattern
             fs, folder_path = url_to_fs(folder_path)
             matched_files = (
                 fs.find(folder_path, detail=False, maxdepth=1 if not recursive else None)
                 if not filename_pattern
-                else fs.glob(filename_pattern.pattern, maxdepth=1 if not recursive else None)
+                else fs.glob(filename_pattern, maxdepth=1 if not recursive else None)
             )
             if not matched_files:
                 raise FileNotFoundError(f'No files matching "{filename_pattern}" found in {folder_path}')
--- a/src/datatrove/utils/dataset.py
+++ b/src/datatrove/utils/dataset.py
@@ -98,7 +98,7 @@ if is_torch_available():
             matched_files = (
                 fs.find(folder_path, detail=False, maxdepth=1 if not recursive else None)
                 if not filename_pattern
-                else fs.glob(filename_pattern, maxdepth=1 if not recursive else None)
+                else fs.glob(filename_pattern.pattern, maxdepth=1 if not recursive else None)
             )
             if not matched_files:
                 raise FileNotFoundError(f'No files matching "{filename_pattern}" found in {folder_path}')
--- a/src/datatrove/utils/dataset.py
+++ b/src/datatrove/utils/dataset.py
@@ -42,12 +42,10 @@ if is_torch_available():
             self._f = None
 
         def __getitem__(self, item):
-            # We loop on the dataset if asking for an index larger than the dataset size
-            epoch_item = item % len(self)
             if not self._f:
                 self._f = self.fs.open(self.file_path, "rb")
             chunk_size = self.token_size * (self.seq_len + 1)
-            self._f.seek(epoch_item * chunk_size)
+            self._f.seek(item * chunk_size)
             return {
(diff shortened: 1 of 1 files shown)


guipenedo:
Indeed you are quite right. The previous version did not use fsspec and thus required a pattern. Let's just make them strings here and remove the `compile`.

> Should we take care of closing the [opened files](https://github.com/huggingface/datatrove/blob/0f2c69f8249aa0c53ebcf10afa2394da506a953f/src/datatrove/utils/dataset.py#L51)?

This would be quite simple to do if we were using [IterableDataset](https://pytorch.org/docs/stable/data.html#iterable-style-datasets), but since these don't allow for indexing the current setup in nanotron and elsewhere would not work great. Unfortunately that also means that there isn't really a clean way to properly close all the files. If you have any ideas I'd be glad to hear them :)

> It's cool to loop back to the beginning of the dataset if we index a position greater than len(dataset) but I can't see when this will happen if both __len__ methods of DatatroveFolderDataset & DatatroveFileDataset return the real length (Used by the sampler for example). Maybe we are hiding indexing positions out of bounds, right?

This is for when you want to epoch on the data, not sure I understand the question on this one

TJ-Solergibert:
The resolution of my 3 comments:
1. `filename_pattern` is of type `str`
2. We take care of closing files with the `__del__` method of `DatatroveFileDataset`. Here is an example:
```python
# Manually override the method to add print message
from datatrove.utils.dataset import DatatroveFileDataset

def del_test(self):
    if self._f:
        print("Closing file!")
        self._f.close()

DatatroveFileDataset.__del__ = del_test
```
```python
from nanotron.data.nanoset import Nanoset
from torch.utils.data import DataLoader

ds = Nanoset(dataset_folders=PATH_TO_PROCESSED_FOLDER,
            sequence_length=2048,
            token_size=2, 
            train_split_num_samples=1000000)

dl = DataLoader(ds, batch_size=1, shuffle=True)

# Consume 10 batches so we open several files
i = 0
for batch in dl:
    i += 1
    if i == 10:
        break

# Delete the dataloader. This is called when the script ends or manually during multistage training in Nanotron, check https://github.com/huggingface/nanotron/blob/67716392ab774a863f1c2f8f73e9571b81ad80a0/src/nanotron/trainer.py#L324
del dl
```
Finally when we override `ds` all the references to `ds` are gone so we `__del__` *everything*. e.g. If we **don't** `del dl` there is still a reference to `ds` so the `__del__` method is not called. 
```python
ds = Nanoset(dataset_folders=PATH_TO_PROCESSED_FOLDER,
            sequence_length=2048,
            token_size=2, 
            train_split_num_samples=5400000)
```
Output:
```
Closing file!
Closing file!
Closing file!
Closing file!
Closing file!
Closing file!
```
3. The user will take care of looping around the dataset if want's to train multiple epochs and NOT the `DatatroveFileDataset` &  `DatatroveFolderDataset` itself.

guipenedo:
Ty :)

PR #210, opened 2024-06-08
Add label_only option to LanguageFilter

justHungryMan:
Follow up # 136

When we set `label_only = True` , this language_filter do not filter but detect language and update language & language_score.

Building on this, I propose we add a new feature to enhance our language analysis capabilities (It's similar stats function with #136 )

The core idea is to introduce a separate block within datatrove dedicated to displaying language statistics. This would not only include the distribution of languages but also provide essential statistics which could be immensely valuable for multi-language processing.

Before moving forward with the development of this feature, I am eager to gather input and insights from the team.

### DIFF
--- a/src/datatrove/pipeline/filters/language_filter.py
+++ b/src/datatrove/pipeline/filters/language_filter.py
@@ -17,6 +17,7 @@ class LanguageFilter(BaseFilter):
         languages: tuple = (Languages.english,),
         language_threshold: float = 0.65,
         exclusion_writer: DiskWriter = None,
+        label_only: bool = False,
     ):
         """
         filters if the predicted language is not among given language or if the language score is below language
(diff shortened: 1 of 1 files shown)


src/datatrove/pipeline/filters/language_filter.py
```diff
@@ -26,10 +27,12 @@ def __init__(
             languages: list of languages to keep
             language_threshold: language_threshold minimum score to accept a document
             exclusion_writer:
+            label_only: if True, only the language label is added to the metadata
```
guipenedo, line 30:
```suggestion
            label_only: if True, only the language label is added to the metadata and no documents are removed
```

guipenedo:
Looks good, thanks!

PR #218, opened 2024-06-13
Fixes text normalization

hynky1999:
Previously the text normalization was extensively destructive due to order of transformation applied.
Eg. `1|2|300$` was transformed to `0`, this means that for many deduplication uses all table lines will be equal.

This PR partially fixes it by both:
1. First applying contextual transformation (multi-char)
2. Converting punctuations to whitespaces instead of directly deleting them.

The new transformation result in `0 0 0`


SinclairCoder:
Hi guys, what effect does this fix cause on dedup? Do you conduct some experiments for that?

guipenedo:
We noticed some non duplicate documents could have matched at the time and investigated. Didn't really conduct in depth ablations

SinclairCoder:
Got it, thanks!

PR #224, opened 2024-06-18
Add shuffle option on huggingface reader

justHungryMan:
Added shuffle option on huggingface reader and also test code for shuffle.

Before merge this commit, plz check the fixed seed value and also buffer size


justHungryMan:
Hi @guipenedo, would you check this PR?

guipenedo:
ty

PR #226, opened 2024-06-18
add alternative fasttext lid models

guipenedo:


src/datatrove/utils/lid.py
```diff
@@ -20,10 +20,11 @@ def predict(self, doc: Document) -> tuple[tuple[str, int], dict[str, float]]:
...
 
 
-class FastTextModel(LID):
-    LANGUAGE_ID_MODEL_URL = "https://dl.fbaipublicfiles.com/fasttext/supervised-models/lid.176.bin"
+class FastTextLID(LID):
+    MODEL_URL = None
+    MODEL_SUBFOLDER = None
 
-    def __init__(self, languages: list[str], k: int = 1) -> None:
+    def __init__(self, languages: list[str], k: int = -1) -> None:
```
hynky1999, line 27: I would allow languages to be nullable in case all languages should be returned

hynky1999 (changes requested):
Small nitpicks

hynky1999 (approved):
LGTM 🚀

PR #229, opened 2024-06-22
DocumentTokenizerContextShuffler fixes

sippycoder:
Fix `.ds.index` file read in shuffler code. Also fix the logging where `path` field is accessed from string.


guipenedo:
LGTM, ty!

PR #230, opened 2024-06-22
Implement zstd Compression Support for JSONL and Parquet Files

justHungryMan:
Support for zstd compression in both JSONL and Parquet file formats. 

Parquet Files:
 - The implementation applies compression directly within the internal write function (pq.ParquetWriter) using the compression option.
 - To maintain clarity and avoid redundancy in DiskWriter implementation, we continue to pass the compression option at the top level but set compression=None in the actual DiskWriter call. This ensures that DiskWriter does not attempt to compress already compressed data.
 - Despite specifying the compression option in ParquetWriter, it is feasible to pass super().__init__(compression=None) to control behavior from a higher abstraction level. This design choice allows flexibility depending on how users might want to integrate or extend our functionality.

Filename Convention:
 - For clarity and standardization, files processed with zstd compression are named with the suffix {file_name}.zstd.parquet (or .zstd.jsonl), aligning with typical conventions and aiding in clear differentiation of compressed files.
 - If the file extension is set to .parquet.zstd, the pq.ParquetFile function fails to properly recognize and read the file. To ensure seamless integration and functionality, the recommended file naming convention is {file_name}.zstd.parquet, which has been tested and confirmed to work effectively with our current setup.

### DIFF
--- a/src/datatrove/pipeline/writers/disk_base.py
+++ b/src/datatrove/pipeline/writers/disk_base.py
@@ -42,6 +42,8 @@ class DiskWriter(PipelineStep, ABC):
         output_filename = output_filename or self.default_output_filename
         if self.compression == "gzip" and not output_filename.endswith(".gz"):
             output_filename += ".gz"
+        elif self.compression == "zstd" and not output_filename.endswith(".zstd"):
+            output_filename += ".zstd"
         self.max_file_size = max_file_size
         self.file_id_counter = Counter()
         if self.max_file_size > 0 and mode != "wb":
--- a/src/datatrove/pipeline/writers/disk_base.py
+++ b/src/datatrove/pipeline/writers/disk_base.py
@@ -40,14 +40,13 @@ class DiskWriter(PipelineStep, ABC):
         self.compression = compression
         self.output_folder = get_datafolder(output_folder)
         output_filename = output_filename or self.default_output_filename
-        if self.compression == "gzip" and not output_filename.endswith(".gz"):
+        if output_filename.endswith(".parquet") and compression != "none":
+            output_filename = output_filename.replace(".parquet", f".{compression}.parquet")
+            compression = None  # compression is already handled by ParquetWriter
+        elif self.compression == "gzip" and not output_filename.endswith(".gz"):
             output_filename += ".gz"
         elif self.compression == "zstd" and not output_filename.endswith(".zstd"):
-            if output_filename.endswith(".parquet"):
-                output_filename = output_filename.replace(".parquet", ".zstd.parquet")
-                compression = None  # compression is already handled by ParquetWriter
-            else:
-                output_filename += ".zstd"
+            output_filename += ".zstd"
         self.max_file_size = max_file_size
         self.file_id_counter = Counter()
         if self.max_file_size > 0 and mode != "wb":
--- a/src/datatrove/pipeline/writers/disk_base.py
+++ b/src/datatrove/pipeline/writers/disk_base.py
@@ -43,7 +43,11 @@ class DiskWriter(PipelineStep, ABC):
         if self.compression == "gzip" and not output_filename.endswith(".gz"):
             output_filename += ".gz"
         elif self.compression == "zstd" and not output_filename.endswith(".zstd"):
-            output_filename += ".zstd"
+            if output_filename.endswith(".parquet"):
+                output_filename = output_filename.replace(".parquet", ".zstd.parquet")
+                compression = None  # compression is already handled by ParquetWriter
+            else:
+                output_filename += ".zstd"
         self.max_file_size = max_file_size
         self.file_id_counter = Counter()
         if self.max_file_size > 0 and mode != "wb":
--- a/src/datatrove/pipeline/writers/parquet.py
+++ b/src/datatrove/pipeline/writers/parquet.py
@@ -1,5 +1,5 @@
 from collections import Counter, defaultdict
-from typing import IO, Callable
+from typing import IO, Callable, Literal
 
 from datatrove.io import DataFolderLike
 from datatrove.pipeline.writers.disk_base import DiskWriter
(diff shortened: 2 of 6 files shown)


src/datatrove/pipeline/writers/disk_base.py
```diff
@@ -40,8 +40,13 @@ def __init__(
         self.compression = compression
         self.output_folder = get_datafolder(output_folder)
         output_filename = output_filename or self.default_output_filename
-        if self.compression == "gzip" and not output_filename.endswith(".gz"):
+        if output_filename.endswith(".parquet") and compression != "none":
```
justHungryMan, line 43: Or we can just use .parquet (not .{compression}.parquet) without adding here.
Then, we do not pass compression option at `src/datatrove/pipeline/writers/parquet.py` line 26.

change to 

```
super().__init__(
    output_folder,
    output_filename,
    compression=None,
    adapter=adapter,
    mode="wb",
    expand_metadata=expand_metadata,
    max_file_size=max_file_size,
)
```

src/datatrove/pipeline/writers/disk_base.py
```diff
@@ -40,8 +40,13 @@ def __init__(
         self.compression = compression
         self.output_folder = get_datafolder(output_folder)
         output_filename = output_filename or self.default_output_filename
-        if self.compression == "gzip" and not output_filename.endswith(".gz"):
+        if output_filename.endswith(".parquet") and compression in ["snappy", "gzip", "brotli", "lz4", "zstd"]:
+            output_filename = output_filename.replace(".parquet", f".{compression}.parquet")
+            compression = None  # compression is already handled by ParquetWriter
+        elif self.compression == "gzip" and not output_filename.endswith(".gz"):
```
guipenedo, line 46: As you suggested I would rather handle this on ParquetWriter directly and pass a `compression=None` to the `super` call
```suggestion
        if self.compression == "gzip" and not output_filename.endswith(".gz"):
```

src/datatrove/pipeline/writers/disk_base.py
```diff
@@ -40,8 +40,13 @@ def __init__(
...
             output_filename += ".gz"
+        elif self.compression == "zstd" and not output_filename.endswith(".zstd"):
+            output_filename += ".zstd"
```
guipenedo, line 49: Isn't .zst more commonly used?
```suggestion
        elif self.compression == "zstd" and not output_filename.endswith(".zst"):
            output_filename += ".zst"
```

src/datatrove/pipeline/writers/parquet.py
```diff
@@ -14,7 +14,7 @@ def __init__(
         self,
         output_folder: DataFolderLike,
         output_filename: str = None,
-        compression: str | None = None,
+        compression: Literal[None, "snappy", "gzip", "brotli", "lz4", "zstd"] | None = None,
```
guipenedo, line 17:
```suggestion
        compression: Literal["snappy", "gzip", "brotli", "lz4", "zstd"] | None = None,
```

src/datatrove/pipeline/writers/parquet.py
```diff
@@ -32,6 +32,7 @@ def __init__(
         self._writers = {}
         self._batches = defaultdict(list)
         self._file_counter = Counter()
+        self.compression = compression
```
guipenedo, line 35: Maybe add an if to check if `compression` is on the list of supported protocols

guipenedo (changes requested):
I think we should go with your idea of handling everything on the ParquetWriter side, by changing https://github.com/huggingface/datatrove/pull/230/files#diff-6b0424e98052b42ca1d50f9fe6008cfb0b4191bbcd33ef7529112db08b5a2b4dR26 to None

src/datatrove/pipeline/writers/disk_base.py
justHungryMan: @guipenedo 
When I used PySpark, files were saved with the .zstd.parquet extension when using zstd compression. However, it seems that .zst is also a commonly used format. I’ve always used .zstd, but I’m not sure which one is more commonly used. 🤨

src/datatrove/pipeline/writers/disk_base.py
justHungryMan: I agree with that.

src/datatrove/pipeline/writers/parquet.py
```diff
@@ -14,24 +14,31 @@ def __init__(
...
         output_folder: DataFolderLike,
         output_filename: str = None,
-        compression: str | None = None,
+        compression: Literal["snappy", "gzip", "brotli", "lz4", "zstd"] | None = None,
         adapter: Callable = None,
         batch_size: int = 1000,
         expand_metadata: bool = False,
         max_file_size: int = 5 * 2**30,  # 5GB
     ):
+        # Validate the compression setting
```
justHungryMan, line 23: add validation check for supported protocols

src/datatrove/pipeline/writers/parquet.py
```diff
@@ -14,24 +14,31 @@ def __init__(
...
+            raise ValueError(
+                "Invalid compression type. Allowed types are 'snappy', 'gzip', 'brotli', 'lz4', 'zstd', or None."
+            )
+
         super().__init__(
             output_folder,
             output_filename,
-            compression,
-            adapter,
+            compression=None,  # Ensure superclass initializes without compression
```
justHungryMan, line 32: Set superclass compression parameter to None to leverage pyarrow's native compression functions.

justHungryMan (commented):
@guipenedo 

Should we include the compression type in the filename within the ParquetWriter code, such as {filename}.snappy.parquet?

src/datatrove/pipeline/writers/disk_base.py
justHungryMan: I think official extension is ".zst" https://datatracker.ietf.org/doc/html/rfc8878#section-7.1

I will fix it

justHungryMan:
HI @guipenedo ,

I used the branch requested in the pull request to process the jsonl.zst file found at [dclm-baseline-1.0](https://huggingface.co/datasets/mlfoundations/dclm-baseline-1.0), and I can confirm that it works correctly. Please review this pull request.

guipenedo (approved):
LGTM, sorry for the delay and thanks a lot!

PR #232, opened 2024-06-28
checks if min_num_sentences is disabled or not before computing the n…

QasidSaleem:
The filter part of C4QualityFilter computes the number of sentences (num_sentences) even if the value of min_num_sentences is set to -1 (disabled), which is unnecessary and incurs extra computational cost.


hynky1999:
lgtm

Issue #239, opened 2024-07-05
need for dependency check for using fasttext


aiqwe:
https://github.com/huggingface/datatrove/blob/061d4db7f26336bd6223af72f282e30db4916121/src/datatrove/utils/lid.py#L40

we need `[fasttext, fasteners]` packages to use fasttext.
so i think it would better to add dependency check function like below.

```python
    @property
    def model(self):
        if not self._model:
            check_required_dependencies("lid", ["fasttext", "fasteners"])
            from fasttext.FastText import _FastText

            model_file = cached_asset_path_or_download(
                self.MODEL_URL,
                namespace="lid",
                subfolder=self.MODEL_SUBFOLDER,
                desc="fast-text language identifier model",
            )
            self._model = _FastText(model_file)
        return self._model
```

guipenedo:
Indeed, good catch. Would you like to make a PR?

aiqwe:
Thank you for your quick reply :D
I made PR #241

PR #241, opened 2024-07-08
add dependencies lid.py, io.py #239

aiqwe:
follow up #239

+ In case of using lid, fasttext is required
+ when we use model property, we use fasteners (in io.py)  

two dependency checks added 😀.

### DIFF
--- a/src/datatrove/utils/lid.py
+++ b/src/datatrove/utils/lid.py
@@ -2,6 +2,7 @@ from abc import abstractmethod
 
 from datatrove.data import Document
 from datatrove.io import cached_asset_path_or_download
+from datatrove.utils._import_utils import check_required_dependencies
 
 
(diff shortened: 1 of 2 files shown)


src/datatrove/utils/lid.py
```diff
@@ -37,6 +38,7 @@ def __init__(self, languages: list[str] | None = None, k: int = -1) -> None:
     @property
     def model(self):
         if not self._model:
+            check_required_dependencies("lid", ["fasttext"])
```
guipenedo, line 41:
```suggestion
            check_required_dependencies("lid", [("fasttext", "fasttext-wheel")])
```

aiqwe (commented):
fixed

guipenedo:
ty!

Issue #249, opened 2024-07-12
Potential issue of dedup in index


jordane95:
Hi, when I'm running the minhash dedup by index, I find the cluster results produced by MinhashDedupCluster is a bit strange.

```
-rw-r--r--    1 root root 108K Jul 12 12:40 001194.clusters
-rw-r--r--    1 root root  54K Jul 12 12:40 001194.remove
-rw-r--r--    1 root root 108K Jul 12 12:40 001195.clusters
-rw-r--r--    1 root root  54K Jul 12 12:40 001195.remove
-rw-r--r--    1 root root 107K Jul 12 12:40 001196.clusters
-rw-r--r--    1 root root  54K Jul 12 12:40 001196.remove
-rw-r--r--    1 root root 107K Jul 12 12:40 001197.clusters
-rw-r--r--    1 root root  54K Jul 12 12:40 001197.remove
-rw-r--r--    1 root root 106K Jul 12 12:40 001198.clusters
-rw-r--r--    1 root root  53K Jul 12 12:40 001198.remove
-rw-r--r--    1 root root 107K Jul 12 12:40 001199.clusters
-rw-r--r--    1 root root  54K Jul 12 12:40 001199.remove
-rw-r--r--    1 root root    8 Jul 12 12:40 4294967295.clusters
-rw-r--r--    1 root root    4 Jul 12 12:40 4294967295.remove
```

There is an outlier which might be due to the SENTINEL token being treated as doc to be removed. So there might be a logical bug in the code?

guipenedo:
Sorry for the (very late) reply. Indeed I think this issue was introduced when we switched merging to be based on set size. If (SENTINEL, SENTINEL) is being marked as a document to be removed, it means that it is not the root of its cluster, and that some other document (the root) that has a match with documents in the index (SENTINEL, SENTINEL) is not being removed.
The point of using the index is removing all the documents present in it, so the solution here instead of that "if" should be  a change in the merge code: when merging two sets, if the root of set A (resp B) is the sentinel then we need to merge B to A (resp A to B), even if the sizes would have us merge the other way

jordane95:
[...]

PR #250, opened 2024-07-13
Fix SENTINEL cluster

jordane95:
We do not want to store cluster_id for sentinel point, since it is not in the current data to process


jordane95:
This should fix #249

PR #251, opened 2024-07-15
Add token and char count to histogram stats

guipenedo:
Adds 2 new stats per original stat when histogram stats are collected:
- `originalname__chars`: number of chars
- `originalname__tokens`: number of tokens (if `metadata["token_count"]` is defined)


hynky1999 (approved):
lgtm

PR #252, opened 2024-07-15
Update MinhashConfig with detailed settings and add default language …

justHungryMan:
While utilizing fuzzy-dedup for Korean documents, I recalled updates related to multi-language tokenization.

**Key Changes:**
- Detailed setting in `MinhashConfig`.
- Explicit default language setting in `MinhashDedupSignature`.

This updates led to the explicit inclusion of a language option in our example code, ensuring that others can easily see how to apply this setting.


hynky1999:
Good idea. I decided to also migrate rest of the examples in this PR.

Thanks!

PR #264, opened 2024-08-08
Fix `test_basic_article_trafilatura` test failure

tylerjthomas9:
"Hello World!" is being duplicated in the `trafilatura` extraction test on `trafilatura==1.12.0` (released last week).

```python
=========================== short test summary info ============================
FAILED tests/pipeline/test_extractors.py::TestExtractors::test_basic_article_trafilatura - AssertionError: 'Hello World!\nHello World!' != 'Hello World!'
- Hello World!
  Hello World!
======= 1 failed, 63 passed, 1 skipped, 19 warnings in 90.94s (0:01:30) ========
```

[This](https://github.com/adbar/trafilatura/issues/660) issue appears relevant. `trafilatura` is duplicating the text because it is too short.

To fix this issue, I modified the config for `trafilatura` to set `MIN_EXTRACTED_SIZE=0`.

I also verified that the issue only appears in `v1.12.0` https://github.com/tylerjthomas9/datatrove/pull/1.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,7 +53,7 @@ processing = [
     "inscriptis",
 #   "readability-lxml @ git+https://github.com/huggingface/python-readability.git@speedup",
     "tldextract",
-    "trafilatura>=1.8.0",
+    "trafilatura>=1.8.0,<1.12.0",
     "tokenizers",
     "ftfy",
     "fasteners",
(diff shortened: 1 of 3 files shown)


hynky1999:
Hi,
I have just tested with `trafilatura==1.12.0` and the test passes.
Could you send here output of:
```
pip freeze
```
?

hynky1999:
Great catch!

I think best way to handle this is to lock the trafilatura to < 1.12.0.
They must have changed the way they process short htmls which is the reason why the duplication happens.
This fix would rather hide it then solve it

PR #271, opened 2024-08-21
Video support for datatrove

guipenedo:
WIP

### DIFF
--- a/src/datatrove/pipeline/filters/video_frozen_filter.py
+++ b/src/datatrove/pipeline/filters/video_frozen_filter.py
@@ -1,0 +1,95 @@
+from typing import Tuple, List
+from datatrove.data import Document
+from datatrove.pipeline.filters.base_filter import BaseFilter
+
+
+class VideoFrozenFilter(BaseFilter):
+    """Filter that uses ffmpeg to detect if a video is static (frozen)."""
+
+    name = "🧊 Video-Frozen-filter"
+    _requires_dependencies = ["ffmpeg"]
+
+    def __init__(self, exclusion_writer=None, batch_size: int = 1, freeze_threshold: float = 0.005, freeze_duration: int = 60):
+        """
+        Args:
+            exclusion_writer: optionally pass in a writer that will save the dropped documents.
+            batch_size: the number of documents to process in a batch.
+            freeze_threshold: the noise threshold for detecting a frozen frame (default is 0.001).
+            freeze_duration: the minimum duration (in seconds) that frames must be frozen to trigger detection (default is 2 seconds).
+        """
+        super().__init__(exclusion_writer, batch_size)
+        self.ffmpeg = None
+        self.freeze_threshold = freeze_threshold
+        self.freeze_duration = freeze_duration
+
+    def filter(self, doc: Document) -> bool | Tuple[bool, str]:
+        video_path = doc.media[0].local_path if doc.media else None
+        import os
+        if not os.path.exists(video_path):
+            print(f"Video path does not exist: {video_path}")
+        if video_path and self.is_video_frozen(video_path):
+            return False, "frozen_video"
+        return True
+
+    def is_video_frozen(self, video_path: str) -> bool:
+        """Dynamically determines intervals and checks if the video is frozen during those intervals."""
+
+        if self.ffmpeg is None:
+            import ffmpeg
+            self.ffmpeg = ffmpeg
+
+        video_duration = self.get_video_duration(video_path)
+        
+        # Adjusted video duration to account for 10-second padding
+        effective_duration = video_duration - 20  # Remove 10 seconds from start and end
+
+        if effective_duration <= 0:
+            # If the effective duration is less than or equal to 0, return False as we can't analyze anything
+            return False
+        
+        intervals = []
+
+        # If the effective duration is very short, analyze the whole effective video
+        if effective_duration < 300:
+            intervals = [("10", str(effective_duration))]
+        else:
+            # Create intervals every 5 minutes (300 seconds), analyzing 1-minute chunks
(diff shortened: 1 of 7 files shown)


src/datatrove/pipeline/readers/videotriplet.py
```diff
@@ -0,0 +1,170 @@
...
+        triplets = []
+        video_extensions = (".mp4", ".avi", ".mkv", ".mov")
+        metadata_extension = ".json"
+        caption_extension = ".vtt"
+
+        if self.paths_file:
+            with self.data_folder.open(self.paths_file, "r") as f:
+                paths = [line.strip() for line in f]
+        else:
+            paths = self.data_folder.list_files(recursive=self.recursive, glob_pattern="*")
```
guipenedo, line 76: if you don't want to use the glob_pattern you can just pass None (or nothing as it is the default), but then you probably want to remove it from the `__init__` args

src/datatrove/pipeline/readers/videotriplet.py
```diff
@@ -0,0 +1,170 @@
...
+
+        for path in paths:
+            base_name, ext = os.path.splitext(path)
+            if ext in video_extensions:
```
guipenedo, line 80: we could maybe extract the triple finding thing away to the datafolder class later on (you'd pass a tuple like `((".mp4", ".avi", ".mkv", ".mov"), ".json", ".vtt")`), but it's probably ok for now

src/datatrove/pipeline/readers/videotriplet.py
```diff
@@ -0,0 +1,170 @@
...
+            default_metadata,
+            recursive,
+            glob_pattern,
+        )
+
+
+    def run(self, data: DocumentsPipeline, rank: int = 0, world_size: int = 1) -> DocumentsPipeline:
+        """Overrides the base run method to handle triplet statistics correctly."""
+        triplet_count = 0
+
```
guipenedo, line 56: This allows stacking multiple readers together (when you want to load data from multiple sources for example)
```suggestion
        if data:
            yield from data
```

src/datatrove/pipeline/readers/videotriplet.py
```diff
@@ -0,0 +1,170 @@
...
+        document = Document(
+            text=caption_text,
+            id=video_id,
+            media=[video_media],
+            metadata=metadata,
+        )
+
+        return document
+
+    def ensure_local_copy(self, video_url: str) -> str:
```
guipenedo, line 125: good job using all of datatrove useful io features here :)

src/datatrove/pipeline/filters/video_frozen_filter.py
```diff
@@ -0,0 +1,95 @@
...
+
+    name = "🧊 Video-Frozen-filter"
+    _requires_dependencies = ["ffmpeg"]
+
+    def __init__(self, exclusion_writer=None, batch_size: int = 1, freeze_threshold: float = 0.005, freeze_duration: int = 60):
+        """
+        Args:
+            exclusion_writer: optionally pass in a writer that will save the dropped documents.
+            batch_size: the number of documents to process in a batch.
+            freeze_threshold: the noise threshold for detecting a frozen frame (default is 0.001).
```
guipenedo, line 17: default don't match the docs

src/datatrove/pipeline/filters/video_frozen_filter.py
```diff
@@ -0,0 +1,95 @@
...
+        super().__init__(exclusion_writer, batch_size)
+        self.ffmpeg = None
+        self.freeze_threshold = freeze_threshold
+        self.freeze_duration = freeze_duration
+
+    def filter(self, doc: Document) -> bool | Tuple[bool, str]:
+        video_path = doc.media[0].local_path if doc.media else None
+        import os
+        if not os.path.exists(video_path):
+            print(f"Video path does not exist: {video_path}")
```
guipenedo, line 29: maybe use `logger.warning` instead (`from loguru import logger`)

src/datatrove/pipeline/filters/video_frozen_filter.py
```diff
@@ -0,0 +1,95 @@
...
+        return False
+
+
+    def get_video_duration(self, video_path: str) -> float:
+        """Get the duration of the video in seconds using ffmpeg."""
+        try:
+            probe = self.ffmpeg.probe(video_path)
+            return float(probe['format']['duration'])
+        except self.ffmpeg.Error as e:
+            print(f"ffprobe {video_path}:")
```
guipenedo, line 77: we usually use `loguru` for all logging

guipenedo (commented):
Not sure how much detail you wanted in the feedback, some stuff are nits but I included them just in case. 
Main issue imo is no sharding on the reader which would be a problem if you want to scale up/use multiple cpus. It's easily fixible, basically just pass `rank` and `world_size` to your find_triplets function and change the return to `return triplets[rank::world_size]`. This should be consistent across workers as `list_files` is deterministic (it sorts the file list) so there should be no overlap between different workers

src/datatrove/pipeline/filters/video_frozen_filter.py
```diff
@@ -0,0 +1,95 @@
+from typing import Tuple, List
+from datatrove.data import Document
+from datatrove.pipeline.filters.base_filter import BaseFilter
+
+
+class VideoFrozenFilter(BaseFilter):
+    """Filter that uses ffmpeg to detect if a video is static (frozen)."""
+
+    name = "🧊 Video-Frozen-filter"
+    _requires_dependencies = ["ffmpeg"]
```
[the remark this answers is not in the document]
guipenedo: You forgot to accept the actual change here I think

guipenedo (commented):
Looking good, just missing the change to check the python lib dependency

PR #286, opened 2024-09-10
Fixed a bug that in the reader pipline, the document count is always less that the actual number of documents by the number of files.

lyuwen:
I came across this bug where I was running comparison of datasets of same number of documents but split into different shards. After inspection of the code, I found that in the code below the number got sent into the documents stat is actually the index of the last docuemnt in a given file, which is less than the number of documents in the file by 1.

https://github.com/huggingface/datatrove/blob/c2fc90213db55afc1c3dcb8d9210b273d9b38d6c/src/datatrove/pipeline/readers/base.py#L193-L204

Therefore, I'm proposing a simple fix that just increment the variable `di` when it's non-zero.

### DIFF
--- a/src/datatrove/pipeline/readers/base.py
+++ b/src/datatrove/pipeline/readers/base.py
@@ -200,6 +200,9 @@ class BaseDiskReader(BaseReader):
                     doc_pbar.update()
                     li += 1
                 file_pbar.update()
+                # document count is non-zero, increment di and store the number
+                # of documents instead of the index of the last document
+                di += min(di, 1)
                 self.stat_update("documents", value=di, unit="input_file")
                 if self.limit != -1 and li >= self.limit:
                     break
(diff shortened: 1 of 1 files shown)


guipenedo:
hi, thanks for the PR
what if there are actually 0 documents? won't this count them as "1" now?
maybe we can just add another variable to count the actual nb of documents? I could see the current version also having issues if we were to use `skip`

lyuwen:
I used the statement
```
di += min(1, di)
```
So that if there were 0 document the stat remains 0.

lyuwen:
Ah, I guess this way my edit would cause a bug when there is exactly 1 document 🤦

lyuwen:
Then I'd suggest to initialize `di` to -1 and always increment `di` after the loop.

lyuwen:
Never mind, an extra variable is need to correctly account for the `skip`.

guipenedo:
LGTM, thanks again!

PR #294, opened 2024-09-29
Fix languages listify bug

BramVanroy:
If the passed language is a string, it should be turned into a list like this `[languages]` rather than `list(languages)`. The latter will turn a string into a list of its individual characters.

```
>>> list("en")
['e', 'n']
>>> ["en"]
['en']
```


hynky1999:
Good catch 👍

hynky1999:
Merging this as tests are failing due to different reason than this PR

PR #296, opened 2024-10-12
[Fixbug] Ensure only one task will be launched for each srun cmd

silverriver:
Some slurm setting will launch multiple tasks for a single srun command. Especially when `--cpus-per-task=1` is used in the sbatch script. 

```bash
$ srun --cpus-per-task=1 --partition=gpu-2080ti echo foo
srun: job 1298286 queued and waiting for resources
srun: job 1298286 has been allocated resources
foo
foo
```

See more detailed discussions [here](https://groups.google.com/g/slurm-users/c/L4nCXtZLlTo)

This behavior is lethal to some tasks if these tasks read and write files. A same set of files will be modified at the same time, which will lead to errors.

It is better to enforce only one task is executed for each srun command.


silverriver:
at @guipenedo to get more attention

hynky1999:
Great catch I wasn't aware this can happen 👍.

PR #304, opened 2024-11-14
Adding Megatron Tokenization pipeline

TJ-Solergibert:
Hi!

In this PR I include the `MegatronDocumentTokenizer` pipeline which stores the tokenized documents in a format compatible with Megatron/NeMo (I'll refer to both projects BUT both produce and require exactly the same files). 

For Megatron/NeMo, we store the document tokens in the same way as `DocumentTokenizer`, in a file with a `.bin` extension. The main difference is that we need to store a `.idx` file which contains some information about the tokenised documents in the `.bin` file & some metadata. The metadata can be a bit tricky to understand but I've included a Byte-Per-Byte explanation of what does it contains. Also, I've dropped support for shuffling and we shouldn't merge the tokenised documents as we should properly handle the `.idx` files. The main part of the code involved in this tokenisation process can be found [here (Megatron)](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/datasets/indexed_dataset.py) & [here (NeMo)](https://github.com/NVIDIA/NeMo/blob/bf7cc64913945ef5e4b02b1c526aa99a912a2153/nemo/collections/nlp/data/language_modeling/megatron/indexed_dataset.py#L380-L625).

I've tested this implementations vs both the [NeMo](https://github.com/NVIDIA/NeMo/blob/main/scripts/nlp_language_modeling/preprocess_data_for_megatron.py) & [Megatron](https://github.com/NVIDIA/Megatron-LM/blob/main/tools/preprocess_data.py) tokenisations scripts (Both are 99% equal, just some minor differences on parallelisation) and we get exactly the same files. I can include some unit tests if necessary, but we would have to port some parts of NeMo/Megatron over here and actively maintain them but it has to be said that this parts of the code haven't been modified for ages.

Toni

### DIFF
--- a/src/datatrove/pipeline/tokens/__init__.py
+++ b/src/datatrove/pipeline/tokens/__init__.py
@@ -1,4 +1,5 @@
 from .context_shuffler import DocumentTokenizerContextShuffler
 from .counter import LengthCounter, TokensCounter
+from .megatron_tokenizer import MegatronDocumentTokenizer
 from .merger import DocumentTokenizerMerger
 from .tokenizer import DocumentTokenizer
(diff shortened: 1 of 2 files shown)


guipenedo:
LGTM, sorry for the wait

PR #307, opened 2024-11-27
[fixbug]: Fixed the issue in MinhashBuildIndex where get_datafolder w…

Youggls:
Fixed the issue in MinhashBuildIndex where get_datafolder was not used to obtain DataFolder for input_folder and output_folder.

In the original code, if a string or tuple type input was passed, an exception would be raised when reading the related files. This was caused by not properly using get_datafolder to obtain the DataFolder type.


guipenedo:
Great catch, thanks!

PR #313, opened 2024-12-11
Add glob pattern for hash index

jordane95:
Avoid irrelevant files when scanning the index folder


guipenedo:
Thanks. Can you run the style fixer? https://github.com/huggingface/datatrove?tab=readme-ov-file#contributing

PR #317, opened 2025-01-06
fix(utils): Enhance the dependencies check to include pip distribution

aiqwe:
With the 0.4.0 release, Datatrove transitioned to numpy 2.0, and the [pyproject.toml ](https://github.com/huggingface/datatrove/blob/2548cdfaacb5e77d1fc4cf1403c8d643c7996090/pyproject.toml#L52)was updated to ensure that the version of fasttext used is compatible with numpy 2.x.

However, if users do not start with a totally new virtual environment and just continue using `fasttext-wheels` compatible with numpy 1.x, we check only verifies the module name `fasttext`, which can cause an [error](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword) to be raised during execution.

To solve this issue, I modfied below:
+ add `_is_distribution_available` in `_import_utils` to strictly manage pip distributions. 
+ modified the `check_required_dependencies` function to perform above.
+ also modified [lid](https://github.com/huggingface/datatrove/blob/v0.4.0/src/datatrove/utils/lid.py#L41), [language_filter](https://github.com/huggingface/datatrove/blob/v0.4.0/src/datatrove/pipeline/filters/language_filter.py#L11), [fasttext_filter](https://github.com/huggingface/datatrove/blob/v0.4.0/src/datatrove/pipeline/filters/fasttext_filter.py) to check pip distribution `fasttext-numpy2-wheel`

During pytest, an error related to the Tibetan language occurred, but since it is not relevant to my region, I did not make any additional changes.

### DIFF
--- a/src/datatrove/pipeline/filters/fasttext_filter.py
+++ b/src/datatrove/pipeline/filters/fasttext_filter.py
@@ -36,7 +36,7 @@ class FastTextClassifierFilter(BaseFilter):
     """
 
     name = "🤖 fastText"
-    _requires_dependencies = [("fasttext", "fasttext-wheel"), "fasteners"]
+    _requires_dependencies = [("fasttext", "fasttext-numpy2-wheel"), "fasteners"]
 
     def __init__(
         self,
--- a/src/datatrove/pipeline/filters/language_filter.py
+++ b/src/datatrove/pipeline/filters/language_filter.py
@@ -8,7 +8,7 @@ from datatrove.utils.lid import FT176LID, GlotLID
 
 class LanguageFilter(BaseFilter):
     name = "🌍 Language ID"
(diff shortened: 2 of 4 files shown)


guipenedo:
Thanks a lot! Can you check the "Allow edits from maintainers" checkbox so that I may fix the tests?
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork

aiqwe:
@guipenedo thx! 😀

However, the checkbox was already checked when I submitted the PR.

스크린샷 2025-01-10 오전 3 05 45

I think the feature could not be working properly.
If necessary, I’m fine with submitting the PR again after the test code is updated.

guipenedo:
my bad, seems to have worked now

PR #327, opened 2025-01-22
Fix issues with URL Deduplication when using the Index

muzzynine:
I found several issues with URL deduplication using the Index and implemented fixes to resolve them.

- Index Signature Ignored: 
The HashSig in the Index must be prioritized to pop first for a given hash value.
To ensure this, the priority of the Index's HashSig should be slightly higher than the maximum priority value of 65,536.

- The Index should be correctly fetched from the index_folder instead of the data_folder.

- When deduplicating a list containing multiple duplicate URLs, the result may still include duplicates.
If the hash_value of `last` matches that of `v`, the update should only occur when `last` is not the HashSig of the Index.
(This issue may also exist in sentence deduplication.)

### DIFF
--- a/src/datatrove/pipeline/dedup/url_dedup.py
+++ b/src/datatrove/pipeline/dedup/url_dedup.py
@@ -51,7 +51,7 @@ class HashSig:
     file_stem: str
 
     def is_from_index(self):
-        return self.doc_id == -1 and self.priority == 1
+        return self.doc_id == -1 and self.priority == 65536
 
     def __lt__(self, other: "HashSig") -> bool:
         # Ensure that highest priority is always first of the hashes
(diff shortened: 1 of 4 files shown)


src/datatrove/pipeline/dedup/url_dedup.py
```diff
@@ -271,12 +271,15 @@ def run(self, data: DocumentsPipeline = None, rank: int = 0, world_size: int = 1
...
+
                 if last and last.hash_value == v.hash_value and not v.is_from_index():
                     out_filename = f"{rank:04d}/{v.file_stem}{ExtensionHelperSD.stage_2_duplicates}"
                     if not index_files or last.is_from_index() or not self.config.only_dedup_in_index:
                         doc_id_bytes = packer.pack(v.doc_id)
                         output_mg.write(out_filename, doc_id_bytes)
-                last = v
+
+                if not last or last.hash_value != v.hash_value or not last.is_from_index():
+                    last = v
```
guipenedo, line 282: good catch for the `only_dedup_in_index=True` case

guipenedo (approved):
Great job, thanks for digging into this.

>  (This issue may also exist in sentence deduplication.)
Indeed it does. Would you mind adding an equivalent if statement for sent dedup in the PR?

muzzynine:
I've added the equivalent `if` statement for sent dedup as requested. 
Let me know if there's anything else you'd like me to adjust. 🤗

guipenedo:
tysm for looking into this :)

PR #330, opened 2025-01-26
Allow custom parquet schema

BramVanroy:
In my data fields I have items that can be None or string, e.g. `doc.metadata["license"] = None`. But if the first document that is written to the file has a None value, parquet will set this field's type to an explicit `null`, so when the writer tries to add a new document that has a non-null value, there will be a conflict.

Explicitly: if the first item is null, the schema will automatically be set it to `pa.null()` but if the next document has a string, it will try using `pa.string()`, which then conflicts with the previous `null()`. To avoid this, we can explicitly pass a schema that sets the type to `pa.string()` to begin with (which allows null), so that the strict `null()` is avoided altogether.

So this PR simply adds the option to pass a Schema to the ParquetWriter (and the HuggingfaceWriter). Default behavior remains the same as before.

### DIFF
--- a/src/datatrove/pipeline/writers/huggingface.py
+++ b/src/datatrove/pipeline/writers/huggingface.py
@@ -2,7 +2,7 @@ import os
 import random
 import tempfile
 import time
-from typing import Callable, Literal
+from typing import Any, Callable, Literal
 
 from huggingface_hub import (
     CommitOperationAdd,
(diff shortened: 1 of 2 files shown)


guipenedo:
Since pyarrow is an optional dependency, the imports need to be local (see the existing pyarrow imports)

BramVanroy:
Fixed - though now the typing is very generic (just `Any`). Not sure if/how that can be improved though.

PR #344, opened 2025-02-12
[draft] Add chunking option to DocumentTokenizer

craffel:


craffel:
@guipenedo I've been using this and it seems to work as intended.

PR #346, opened 2025-02-13
fix bos token missing

### LINKED ISSUE #345: BOS tokens not properly added in some circumstances
When using `DocumentTokenizer` if an `eos_token` is specified the tokenizer post processor is replaced with one that appends and EOS. However, this has the effect of NOT placing a BOS token at the beginning of the sequence.

See here: https://github.com/huggingface/datatrove/blob/main/src/datatrove/utils/tokenization.py#L55

This can be reproduced by tokenizing with a tokenizer like Llama 3 and looking at the raw token values

jquesnelle:
This fixes #345 by checking for all three combinations of bos/eos existence and using an appropriate template for each case


guipenedo:
hi, can you please make this pr editable by maintainers?

PR #349, opened 2025-03-04
fix: root condition for SENTINEL

jordane95:
This should fix https://github.com/huggingface/datatrove/issues/249.


guipenedo:
What if root_b is the sentinel and size_b < size_a?

jordane95:
> What if root_b is the sentinel and size_b < size_a?

thx for correction, add one more condition for this case

PR #361, opened 2025-04-24
add oom score + shorter polling

hynky1999:


src/datatrove/pipeline/extractors/base.py
```diff
@@ -114,13 +127,28 @@ def process_document(self, text, extract_fn):
...
-                return result
-            else:
-                raise TimeoutError("Document extraction timed out")
+
+            deadline = time.time() + self.timeout
+            # loop with short sleeps instead of one big poll()
+            while True:
+                # 5 seconds is small enough not to take cpu time too much but not big enough so that we return quikly
+                # on child process death, so that we indeed after the return (not sure if needed)
+                poll_timeout = min(5, deadline - time.time() + 0.1)
```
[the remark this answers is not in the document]
hynky1999: Very smart mr copilot

src/datatrove/pipeline/extractors/base.py
```diff
@@ -114,13 +127,28 @@ def process_document(self, text, extract_fn):
...
             self.parent_conn.send(text)
-            if self.parent_conn.poll(timeout=self.timeout):
-                result = self.parent_conn.recv()
-                if isinstance(result, Exception):
-                    raise result
-                return result
-            else:
-                raise TimeoutError("Document extraction timed out")
+
+            deadline = time.time() + self.timeout
```
guipenedo, line 131:
```suggestion
            deadline = time.monotonic() + self.timeout
```
time.monotonic is often used for timeouts, better resolution than time.time

src/datatrove/pipeline/extractors/base.py
```diff
@@ -114,13 +127,28 @@ def process_document(self, text, extract_fn):
...
+            # loop with short sleeps instead of one big poll()
+            while True:
+                # 5 seconds is small enough not to take cpu time too much but not big enough so that we return quikly
+                # on child process death, so that we indeed after the return (not sure if needed)
+                poll_timeout = max(0, min(5, deadline - time.time() + 0.1))
```
guipenedo, line 136:
```suggestion
                poll_timeout = max(0, min(5, deadline - time.monotonic() + 0.1))
```
same as above

src/datatrove/pipeline/extractors/base.py
```diff
@@ -114,13 +127,28 @@ def process_document(self, text, extract_fn):
...
+                        raise result
+                    return result
+
+                # 2) Has the child died?
+                if not self.process.is_alive():
+                    self._cleanup_process()
+                    raise EOFError("Child process died (likely OOM-killed)")
+
+                # 3) Has our deadline passed?
+                if time.time() >= deadline:
```
guipenedo, line 149:
```suggestion
                if time.monotonic() >= deadline:
```

src/datatrove/pipeline/extractors/base.py
```diff
@@ -114,13 +127,28 @@ def process_document(self, text, extract_fn):
...
-                result = self.parent_conn.recv()
-                if isinstance(result, Exception):
-                    raise result
-                return result
-            else:
-                raise TimeoutError("Document extraction timed out")
+
+            deadline = time.time() + self.timeout
+            # loop with short sleeps instead of one big poll()
+            while True:
```
guipenedo, line 133: do we actually need this to be in a loop? does it not work otherwise? because if the reason it works is having a high enough polling interval this won't work for the shorter intervals I normally use (1-2s)

guipenedo (commented):
do we actually need this to be in a loop? does it not work otherwise? because if the reason it works is having a high enough polling interval this won't work for the shorter intervals I normally use (1-2s)

src/datatrove/pipeline/extractors/base.py
hynky1999: Yes we do, because if you timeout is 1 minutes. You want to check every 5 seconds for subprocess disconnect. This works with your short timeout just fine because: `poll_timeout = max(0, min(5, deadline - time.monotonic() + 0.1))` This will essentially set it to timeout if timeout < 5. Then you do just one cycle :)

PR #366, opened 2025-05-08
ensure folder_path has consistent usage

hynky1999:


src/datatrove/utils/dataset.py
```diff
@@ -278,15 +278,19 @@ def __init__(
...
             fsizes = {}
 
             if not self.paths_file or not file_exists(self.paths_file):
-                matched_files = self.folder_path.list_files(
+                matched_files = self.data_folder.list_files(
                     glob_pattern=self.filename_pattern, recursive=self.recursive
                 )
                 if not matched_files:
-                    raise FileNotFoundError(f'No files matching "{self.filename_pattern}" found in {self.folder_path}')
+                    raise FileNotFoundError(f'No files matching "{self.filename_pattern}" found in {self.data_folder}')
```
guipenedo, line 293: here should be self.folder_path (string)

guipenedo:
just added a note for the exception, feel free to merge otherwise

PR #371, opened 2025-06-25
update ray doc

Tavish9:
This PR update README.md about `ray`'s doc


Tavish9:
I think we can always add `ray.init()` whether the cluster is set-up or not.

`ray.init()` would init a ray cluster if no cluster is available, or join a existing cluster.

hynky1999:
> I think we can always add ray.init() whether the cluster is set-up or not.
Yeah, that's what i meant
Thanks :)

PR #385, opened 2025-08-29
Add support to load HF dataset from disk

iamgroot42:
Allow option in `HuggingFaceDatasetReader` to read a Huggingface dataset stored locally


iamgroot42:
@guipenedo  bumping this up to see if it might be worth adding?

guipenedo:
LGTM, sorry for the delay

PR #398, opened 2025-11-07
Inference runner refactoring: rollouts, gen params, etc

guipenedo:
Heavy refactor of the InferenceRunner:
- temperature moved inside the new `default_generation_params` InferenceConfig option. These parameters are automatically added to request payloads if not present (the user can override them by setting them explicitly on the payload)
- changed the general API from `query_builder` (returns query payloads), to `rollout_fn`: receives a `generate` callback that it can call multiple times per rollout (useful for tool calling or other multi-step approaches)
- new `rollouts_per_document` setting: the user defined `rollout_fn` is called this many times and results are automatically aggregated and saved in `metadata['rollout_results']` 
- simplified InferenceSuccess, InferenceError, InferenceProcessingError to simply `InferenceResult` and `InferenceError` (an exception)
- returning `None` from all rollouts will drop the document (it will not be saved to the output)
- request caching using a sqlite database: for incomplete documents, we can now instantly replay previously executed requests. Useful for long multi step workflows
- improved the rollouts api: no longer passing `runner` around, and we now allow users to send a `shared_context` that is passed to rollouts (either a dictionary, a function returning a dictionary, or a context manager that yields a dictionary)
- added EndpointServer: simpler version that runs for a specific endpoint without spawning an inference server. Uses raw sockets if endpoint is local and the openai package otherwise.

To be added in later PRs if needed:
- multi-node deployments
- extending request caching to tools

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,13 +321,57 @@ Some options common to most readers:
 - `limit` read only a certain number of samples. Useful for testing/debugging
 
 ### Synthetic data generation
-We support [vLLM](https://github.com/vllm-project/vllm) and [SGLang](https://github.com/sgl-project/sglang) for inference using the [InferenceRunner block](src/datatrove/pipeline/inference/run_inference.py). Each datatrove task will spawn a replica of the target model and full asynchronous continuous batching will guarantee high model throughput.
+Install the inference extras with `pip install datatrove[inference]` to pull in the lightweight HTTP client, checkpointing dependencies and async sqlite cache.
 
-By setting `checkpoints_local_dir` and `records_per_chunk` generations will be written to a local folder until a chunk is complete, allowing for checkpointing in case tasks fail or are preempted.
+We support [vLLM](https://github.com/vllm-project/vllm), [SGLang](https://github.com/sgl-project/sglang), OpenAI-compatible HTTPS endpoints and a local `dummy` server through the [InferenceRunner block](src/datatrove/pipeline/inference/run_inference.py). Each datatrove task can spin up its own server replica (for `vllm`, `sglang` or `dummy`) or talk directly to an external endpoint while asynchronous batching keeps GPU utilization high.
 
-Tune `max_concurrent_requests` to tune batching behaviour. If you have slow pre-processing, you can also increase `max_concurrent_tasks` (to a value higher than `max_concurrent_requests`). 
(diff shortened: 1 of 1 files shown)


src/datatrove/pipeline/inference/run_inference.py
```diff
@@ -637,78 +624,32 @@ async def _handle_record(
...
-                    postprocess_result = self.postprocess_fn(self, doc)
-                    if postprocess_result is None:
-                        doc.metadata["postprocess_remove"] = True
-                    else:
-                        doc = postprocess_result
-
-                if callback_requested:
-                    await _handle_record(doc, rank, chunk_index, output_writer_context)
-                    return
+                generate_callback = partial(self._send_request, semaphore=semaphore)
```
hynky1999: No need to  be in try block and overall this object needs to be created just once when we are initializing the semaphore. Move it out of the fc and then just pass the generate_callback instead of the semaphore

guipenedo: > Move it out of the fc and then just pass the generate_callback instead of the semaphore

you're right if it stays as is, but for the cache I will make this function receive doc_id and rollout_idx, so I will need to call partial each time anyway

> I don't like the fact that the InferenceError raise from any rollout cause all rollouts to fail. the try/except should be applied per rollout not on all of them. Also is that intended that any inference error causes whole thing to die ?

the idea is that the user should be the one handling generation errors now since they call the function. If they don't, the whole thing dies

src/datatrove/pipeline/inference/run_inference.py
```diff
@@ -140,249 +105,100 @@ class InferenceConfig:
...
+    # metrics
     metric_interval: int = 120
+    # parallelism
     tp: int = 1
     dp: int = 1
     pp: int = 1
-    max_concurrent_tasks: int | None = None
-    use_chat: bool = True
+    # rollouts and generation
+    default_generation_params: dict = field(default_factory=lambda: {"temperature": 0.0})
```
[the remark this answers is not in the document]
guipenedo: I will remove the default value of temperature 0.0

src/datatrove/pipeline/inference/run_inference.py
```diff
@@ -637,126 +521,110 @@ async def _handle_record(
...
 
                 await self._save_document(doc, output_writer_context, rank, chunk_index)
-            except InferenceProcessingError as e:
+                await self.request_cache.mark_document_complete(doc.id)
+            except InferenceError as e:
                 raise e
             except Exception as e:
                 # let's propagate it
-                raise InferenceProcessingError(doc, e)
+                raise InferenceError(doc, e)
```
hynky1999: The constructor is using incorrect args

PR #407, opened 2025-11-24
Fix missing param

shallyan:
ExtractorSandbox requires the param "wamup_text", but it is not given, which will cause the failure of local execution.


hynky1999:
Thanks :)

PR #408, opened 2025-12-08
Streaming HF Dataset Writer

JoelNiklaus:
The `HuggingFaceDatasetWriter` only uploads the data at the end. When we want to run inference on a large dataset, we may want to continuously upload data to the hub so we can monitor progress. We implement this in [DataForge](https://github.com/huggingface/dataforge/pull/31). This PR is for upstreaming this writer to datatrove.

### DIFF
--- a/src/datatrove/pipeline/writers/huggingface.py
+++ b/src/datatrove/pipeline/writers/huggingface.py
@@ -215,9 +215,9 @@ class StreamingHuggingFaceDatasetWriter(HuggingFaceDatasetWriter):
                 break
             except HfHubHTTPError as exc:
                 race_condition = False
-                if hasattr(exc, 'server_message'):
+                if hasattr(exc, "server_message"):
                     race_condition = "A commit has happened since" in exc.server_message
-                elif hasattr(exc, 'response') and exc.response is not None:
+                elif hasattr(exc, "response") and exc.response is not None:
                     race_condition = "A commit has happened since" in exc.response.text
                 if race_condition and retries < MAX_RETRIES:
                     logger.warning(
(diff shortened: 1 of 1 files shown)


JoelNiklaus:
I have tried the `HuggingFaceDatasetWriter` which inherits from `ParquetWriter`. Unfortunately, it does not seem to stream. It seems to upload everything in the end only.

It stages files to LFS during the run but defers the actual commit (making them visible in the repo) until close() is called. Each call to `upload_files()` pushes bytes to Hugging Face's LFS backend via `preupload_lfs_files()` and appends to an internal operations list, yet the single `create_commit()` that finalizes those operations only fires at teardown. This design minimizes commit noise and avoids race conditions for typical batch jobs, but it means external observers (like a progress monitor) see zero documents until the entire job finishes.


```
# From datatrove/pipeline/writers/huggingface.py

def upload_files(self, *filenames):
    # ...
    preupload_lfs_files(self.dataset, repo_type="dataset", additions=additions, revision=self.revision)
    # ...
    self.operations.extend(additions)  # staged, not committed

def close(self, rank: int = 0):
    # ...
    create_commit(                      # <-- only here do files become visible
        self.dataset,
        repo_type="dataset",
        operations=self.operations,
        commit_message=f"DataTrove upload ({len(self.operations)} files)",
        revision=self.revision,
    )
```

guipenedo:
This is fully intended, the hub has a hard time handling many commits and that block was created to upload large (complete) datasets efficiently. I don't guarantee it but I think plain ParquetWriter with a hf:// output path should work as you intended. If it doesn't we can add this new block

PR #410, opened 2025-12-19
feat: Add full inference server response to `InferenceResult`

hynky1999:


guipenedo:
For proper typing, something like:
```
from dataclasses import dataclass, field
from typing import Any, Dict, List, Literal, Optional, Union


Role = Literal["system", "user", "assistant", "tool"]


@dataclass
class ToolCall:
    id: Optional[str] = None
    name: Optional[str] = None
    arguments: Dict[str, Any] = field(default_factory=dict)


ContentPart = Union[
    str,                              # plain text
    Dict[str, Any],                   # image/audio/etc blocks
]


@dataclass
class Message:
    role: Role

    # Main user-visible output
    content: Optional[Union[str, List[ContentPart]]] = None

    # Tool / function calling
    tool_calls: List[ToolCall] = field(default_factory=list)
    tool_call_id: Optional[str] = None   # used on tool responses

    # Optional metadata (model/provider specific)
    reasoning: Optional[str] = None
    name: Optional[str] = None           # function name / tool name
    metadata: Dict[str, Any] = field(default_factory=dict)
```

hynky1999:
> For proper typing, something like:

I am not sure it's good idea to add typing. To me the only reason we are adding the raw response is to enable users to extract stuff on top of what we already parse so we should be as flexible as possible there so I would just keep it as is.

PR #493, opened 2026-06-28
refactor(filters): use Counter(x) fast path in GopherRepetitionFilter

mattfaltyn:
## Summary

`GopherRepetitionFilter` runs on every document in common pipelines (FineWeb, Gopher). Its `find_top_duplicate` helper is called 3x per document (n=2,3,4) and used a manual per-element increment loop:

```python
counter = Counter()
for element in x:
    counter[element] += 1
```

Replaced with `Counter(x)`, which uses CPython's optimized C counting path. This is **behavior-preserving**: both insert keys in first-seen order and `most_common` is a stable sort, so tie-breaking is identical.

## Benchmark (mock data)

Mock documents with a Zipf-ish vocabulary (so common words repeat and the duplicate path is exercised). Timings are the min of 5 runs over 200 docs; CPython 3.13, Apple Silicon. Reproduction script is in the PR thread.

| words/doc | old (loop) | new (`Counter(x)`) | speedup |
|---:|---:|---:|---:|
| 200 | 20.0 us | 7.1 us | **2.82x** |
| 1000 | 99.3 us | 34.2 us | **2.90x** |
| 5000 | 494.5 us | 170.3 us | **2.90x** |

## Tests

Added a `TestGopherRepetitionHelpers` suite in `tests/pipeline/filters/test_filters.py` (the helper operates on a list of strings, so no `nltk`/`spacy` needed):

- `test_find_top_duplicate` — golden values incl. tie-breaking and single-element.
- `test_find_top_duplicate_empty_raises` — documents existing empty-input behavior (the filter guards with `if not n_grams`).
- `test_find_top_duplicate_differential` — asserts `Counter(x)` matches the original manual loop over **1000 randomized inputs** (seeded `random`, no new dependency).

## Test plan

- [x] `pytest tests/pipeline/filters/test_filters.py::TestGopherRepetitionHelpers` — 3 passed
- [x] `ruff check` + `ruff format --check` on changed files — clean
- [x] Differential equivalence (new == old) verified on 1000 randomized inputs
- [ ] Full `make test` in CI (existing `test_gopher_repetition` routes through this helper and requires `spacy`/`nltk`)

### DIFF
--- a/src/datatrove/pipeline/filters/gopher_repetition_filter.py
+++ b/src/datatrove/pipeline/filters/gopher_repetition_filter.py
@@ -47,9 +47,7 @@ def find_duplicates(x: list[str]) -> tuple[int, int]:
 
 
 def find_top_duplicate(x: list[str]) -> int:
-    counter = Counter()
-    for element in x:
-        counter[element] += 1
+    counter = Counter(x)
     top_n_gram = counter.most_common(1)[0]
     return len(top_n_gram[0]) * top_n_gram[1]
 
--- a/tests/pipeline/filters/test_filters.py
+++ b/tests/pipeline/filters/test_filters.py
@@ -1,4 +1,6 @@
+import random
 import unittest
+from collections import Counter
 
 from datatrove.data import Document
 from datatrove.pipeline.filters import (
@@ -12,6 +14,7 @@ from datatrove.pipeline.filters import (
 )
 from datatrove.pipeline.filters.c4_filters import C4ParagraphFilter, C4QualityFilter
 from datatrove.pipeline.filters.fineweb_quality_filter import FineWebQualityFilter
+from datatrove.pipeline.filters.gopher_repetition_filter import find_top_duplicate
 from datatrove.pipeline.filters.sampler_filter import SamplerFilter
 
 from ...utils import require_fasttext, require_nltk, require_tldextract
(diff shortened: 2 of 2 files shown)


mattfaltyn:
<details>
<summary>Benchmark reproduction script</summary>

```python
import random, string, timeit
from collections import Counter

def old_find_top_duplicate(x):
    counter = Counter()
    for e in x:
        counter[e] += 1
    top = counter.most_common(1)[0]
    return len(top[0]) * top[1]

def new_find_top_duplicate(x):
    top = Counter(x).most_common(1)[0]
    return len(top[0]) * top[1]

def make_doc(n_words, seed):
    rng = random.Random(seed)
    vocab = ["".join(rng.choices(string.ascii_lowercase, k=rng.randint(2, 10)))
             for _ in range(max(50, n_words // 5))]
    weights = [1.0 / (i + 1) for i in range(len(vocab))]
    return rng.choices(vocab, weights=weights, k=n_words)

def get_n_grams(words, n):
    return [" ".join(words[i:i+n]) for i in range(len(words) - n + 1)]

for size in (200, 1000, 5000):
    grams = [get_n_grams(make_doc(size, i), 3) for i in range(200)]
    old = min(timeit.repeat(lambda: [old_find_top_duplicate(g) for g in grams], number=1, repeat=5))
    new = min(timeit.repeat(lambda: [new_find_top_duplicate(g) for g in grams], number=1, repeat=5))
    print(f"{size:>5} words/doc | old {old/200*1e6:7.1f} us | new {new/200*1e6:7.1f} us | {old/new:.2f}x")
```

Note: I also benchmarked removing the `find_all_duplicate` assert entirely, which is ~1.05-1.07x faster than the original. I kept the assert (as the cheaper `sum()` form) to preserve the existing sanity check while still eliminating the per-call document-sized allocation; happy to drop it instead if maintainers prefer.

mattfaltyn:
Update: I narrowed this PR to **only** the `find_top_duplicate` -> `Counter(x)` change (the clear ~2.9x win), and reverted the `find_all_duplicate` assert tweak I mentioned above. That assert change was CPU-neutral (it was really about avoiding a per-call allocation), so it's better left for a separate, dedicated discussion rather than muddying this one. The reproduction script above remains valid for the `Counter` benchmark.

mattfaltyn:
Needs more review

PR #497, opened 2026-07-07
Add JobsPipelineExecutor: run datatrove pipelines on HF Jobs (experimental)

davanstrien:
**Framing.** Started as a draft proposal; after discussion (thanks @lvwerra, @guipenedo) this is now intended to land as an **explicitly experimental** feature: the executor warns on init, the docstrings say the API may change or break, and none of the core datatrove logic is touched (the diff is additive: one new executor + examples + tests). Claude Code did most of the heavy lifting after discussing the design with me.

### Review round (July 10–14)
Independently verified by @JoelNiklaus (all three examples as real Jobs + the unit suite). All feedback addressed: auto-create of the `hf://` logging repo/bucket (ea0661d, with tests), examples' `DEPENDENCIES` literally install from git + minhash example defaults to a Storage Bucket (18ae3f4), fs-cache refresh on the launch path (9f41882, via Bugbot). **Until this ships in a release, install with:** `datatrove[io,processing] @ git+https://github.com/huggingface/datatrove`.

### What's here
- `JobsPipelineExecutor(PipelineExecutor)` in `executor/jobs.py` — mirrors `SlurmPipelineExecutor` but launches Jobs via `huggingface_hub.run_uv_job` instead of `sbatch`.
- 3 adapted examples (`filter_hf_dataset_jobs.py`, `tokenize_hf_dataset_jobs.py`, `minhash_deduplication_jobs.py`) + unit tests (mock the Jobs API, `memory://` logging_dir).

### Review pass before undrafting (July 9)
Two independent reviews of the full diff (Codex + Claude), fixes applied for everything high-confidence, each pinned by a new test that fails on the previous code:
- **Credential scrub (the important one):** `token=` / `secrets=` passed to the executor were dill-pickled into `executor.pik` and serialized into `executor.json` on the shared (possibly public) `logging_dir`. They are now scrubbed before anything is persisted — Jobs receive credentials exclusively as Job secrets.
- Stats: `stats.json` is now merged when `skip_completed=False`, and re-merged on rerun if a previous coordinator died after the last rank completed but before the merge.
- `tasks_per_job` / `workers` are validated (`workers=0` used to silently launch nothing; `tasks_per_job=0` was a raw `ZeroDivisionError`).
- `randomize_start_duration` was documented but dropped by the `_run_for_rank` override — now honored (mirrors the base executor).
- Docstring: cancel stale Jobs from a previous launch before rerunning (each launch rewrites `executor.pik` / `ranks_to_run.json`).

### Verified on real Jobs (July 9, this branch)
- Full CI-equivalent test suite (`uv sync --extra testing` + `pytest ./tests/`) run in a cpu-upgrade Job: **381 passed, 4 skipped** (test_ray excluded — Ray segfaults in Job containers regardless of this branch; covered by GitHub CI).
- Single-stage fan-out (filter example): 4 Jobs, 100k docs read, filtered output written, stats merged; experimental warning is the first coordinator line.
- 4-stage minhash `depends=` chain at realistic scale (ag_news, 120k docs, 24 signature tasks). Two findings on `hf://` *dataset repos* as the shared store: (1) at 8 concurrent Jobs committing signature files to one repo, 6/24 ranks exhausted retries on Hub commit rate limits (429 "maximum time in concurrency queue reached") — the chain **failed fast with resume guidance as designed**, and rerunning at `workers=4` resumed exactly the 6 failed ranks and completed stage 1; (2) stage-2 bucket-merge tasks (a 24-way priority-queue merge over remote signature files, i.e. many tiny ranged reads) run ~45 min each over a dataset repo — slow but correct, and fixed by `lines_to_buffer=-1` (see the follow-up comment: 65 min → 99 s). The full 4-stage chain completed end-to-end: 120,000 docs in → 115,736 kept, 4,264 near-duplicates removed.
- **Buckets answer the consistency question (Q4) — for writes:** the same signature stage on a Storage Bucket (`hf://buckets/...`, same HfFileSystem code path, no code changes, needs huggingface_hub>=1.6.0) passed **24/24 at `workers=8` with zero retries** — no commit queue, no read-after-write lag. Caveat from the same runs: buckets do **not** rescue the merge stage — stage-2 tasks take ~45 min (repo) vs ~30 min (bucket) per task, the same order of magnitude, because the dominant cost is thousands of tiny ranged reads through `HfFileSystem` (Slurm avoids this with node-local disk; a fix would be read buffering/local staging, not a store choice). Both findings are documented in the minhash example.
- Version floor sanity check: `run_uv_job`/`inspect_job` import cleanly at exactly `huggingface-hub==1.5.0` (the current pin); `hf://buckets` paths need >=1.6.0 (noted in the docstring).
- Failure semantics for real: a stage with deterministically-failing ranks exhausts `max_retries` (resubmit observed), the dependent stage fails fast with `RuntimeError` (`run_on_dependency_fail=False`); rerunning with a fixed pipeline relaunched **only** the 4 previously-failed ranks (idempotent resume), after which the child ran clean.
- 2-stage tokenize→merge (16 fan-out tasks + 1 fan-in task): completed, stats merged for both stages.
- Credential scrub verified against a real `hf://` logging repo: ran with an explicit `token=` + canary secret; neither is present in the persisted `executor.pik` / `executor.json`, and the Jobs still authenticate (via Job secrets).

### Known limitations (accepted for an experimental first cut)
- `run()` does not raise when ranks remain permanently failed after retries — it logs and leaves the completion markers for an idempotent rerun (fire-and-forget-ish, like Slurm). A dependent stage *does* fail fast. Open question whether `run()` itself should raise.
- A child whose `depends=` parent was run in a *different* process and left permanently incomplete keeps waiting (2-min polls) — cross-process, markers alone can't distinguish "still running" from "failed for good".
- Coordinator calls to `run_uv_job` / `inspect_job` have no retry/backoff — a transient API error kills the coordinator (in-flight Jobs finish; rerun resumes).
- `timeout=None` (default) has no coordinator-side deadline for a Job wedged in a non-terminal state; set `timeout=` for an upper bound.

### How it works (the Slurm→Jobs mapping)
Reuses the Slurm machinery unchanged — dill-pickle the executor to `logging_dir/executor.pik`, write `ranks_to_run.json`, reuse the existing `launch_pickled_pipeline` worker. Three swaps: (1) `SLURM_ARRAY_TASK_ID` → `DATATROVE_JOB_ARRAY_INDEX`; (2) `sbatch --array 0-M%W` → K `run_uv_job` calls throttled by a local polling loop; (3) `--dependency=afterok` → poll `completions/`. Same `executor.json`/`completions/` layout, so `jobs_status`/`track_jobs` work unchanged.

### Open design questions (for post-merge iteration)

1. **Coordination model.** The launcher currently runs *locally* and blocks for the whole run — it submits Jobs, polls to honor `workers`, and merges stats (like Local/Ray; only Slurm is fire-and-forget). Is that the model you want, or should the canonical path be a **coordinator Job** (runs the launcher itself so nothing stays on the client), or built on **`SandboxPool`** (packs workers onto warm host VMs)?
   *Lean: the local launcher is the simplest first cut (what's here); a coordinator Job is the obvious way to free the client; `SandboxPool` is the interesting option for many tiny tasks (CPU-only, bigger change). The "don't launch Jobs from within a Job" note is about the worker path — a dedicated coordinator Job is a separate role.*

2. **Worker entrypoint / getting datatrove onto the worker.** Current: `run_uv_job` uploads `launch_pickled_pipeline` and installs datatrove into a fresh `uv` env via `dependencies` — a ~30–90s cold install per Job (worse with spacy/torch). But Jobs accept an arbitrary image (`run_job(image=…)` takes a Docker ref, or even an `hf.co/spaces/…` image), so the entrypoint can work several ways: uv-install-per-job (current), a prebuilt / uv-baked datatrove image (warm start), or a Space-as-image. Which do you want as the default — and is a maintained datatrove base image something you'd want to own?
   *Lean: uv-install is a fine default (nothing to maintain; the install is noise against a real job's runtime); an optional `image=` gives the warm-start path (prebuilt / uv-baked / Space). A blessed base image only if you'd want to own it. Observed today: total per-Job overhead (scheduling + uv env build with `datatrove[io,processing]`) is ~1–2 min on cpu-basic — so size tasks to at least several minutes of work each.*

3. **Dependency declaration.** Users must currently list every extra their steps need, and it's non-obvious (even `LambdaFilter` needs `[processing]` because `datatrove.pipeline.filters` hard-imports `regex`; minhash needs bare `spacy`). Leave it manual, or infer from steps' `_requires_dependencies`? (Inference catches `spacy` but *not* `regex` — the latter is a module-level import, not a declared dep.) Related: datatrove version pinning for pickle-compat is currently the user's responsibility via the dep string — should the executor pin it?

4. **Store & consistency.** `logging_dir` must be remote. On `hf://` we hit real issues — read-after-write 404s between stages and client-side listing-cache staleness (see Issues); patched enough to work (retries + cache invalidation) but rough. A strongly-consistent store — e.g. HF's **S3-compatible buckets** (closer to how datatrove already uses `s3://`) — would likely be a cleaner fit than `hf://` for multi-stage chains. Open question, not something we've solved.

5. **`depends=` semantics.** Chaining is blocking recursion (launch parent → block → run child). We added fail-fast on a permanently-failed dependency + a `run_on_dependency_fail` flag (mirrors `afterok`/`afterany`). Is that the semantics you want, or do you want a real DAG/scheduler rather than nested blocking launches?

6. **Concurrency semantics of `workers`.** It's a local cap enforced by polling `inspect_job` (no native Jobs array/`%W`). On failure we resubmit up to `max_retries`. Right knobs/defaults?

7. **Scope.** Kept text-only (matches the rest of datatrove). Design the worker contract with media in mind now, or defer?

### Issues hit (and what we did)
- **Per-rank stat inflation** when `tasks_per_job>1` — ranks shared pipeline-step objects, so stats accumulated. Deepcopy the pipeline per rank (note: `slurm.py` has this latent too — data was always fine, only stats over-counted).
- **`depends=` hung forever** — the launcher's `HfFileSystem` cached `completions/` and never refreshed, so it couldn't see completions the *remote Jobs* wrote (fresh process saw 14/14; launcher saw 13/14). Fixed with `invalidate_cache()` before each poll + stats merge. This is the one that would bite anyone using `depends=` on `hf://`.
- **Infinite wait on a failed dependency** — poll loop waited forever; added fail-fast + opt-out.
- **`hf://` read-after-write 404** between stages — `max_retries≥1` covers it (see Q4).
- **Dependency gotchas** — `[processing]` for filters (`regex`), bare `spacy` for the English tokenizer (only shipped in the heavy `multilingual` extra).
- **uv cold-install per Job** — perf cost, see Q2.

### Alternative paths considered / not taken
- **Pickle a stock `LocalPipelineExecutor` instead of `self`** → workers could run stock PyPI datatrove (no fork install). Rejected for Slurm-symmetry, but a real option if you'd rather not require datatrove-on-the-worker to match.
- **`run_job` + prebuilt/uv-baked image** instead of `run_uv_job` (Q2).
- **`SandboxPool` backend** — CPU-only, but amortizes install across a warm host pool; likely the better long-term shape for many small tasks (Q1).

---

> [!NOTE]
> **Medium Risk**
> New distributed execution path touches remote storage coordination, credential handling, and multi-stage `depends=` semantics; core pipeline code is unchanged but misconfigured `logging_dir`/deps or stale Jobs could cause failed or partial runs.
> 
> **Overview**
> Adds an **experimental** `JobsPipelineExecutor` so datatrove pipelines can run on [Hugging Face Jobs](https://huggingface.co/docs/huggingface_hub/en/guides/jobs) with the same coordination model as Slurm: pickle the executor to a shared remote `logging_dir`, shard work via `ranks_to_run.json` and `completions/`, and resume by skipping finished ranks.
> 
> The launcher blocks locally, submits Jobs through `huggingface_hub.run_uv_job` (each runs `launch_pickled_pipeline` in a fresh `uv` env from `dependencies`), and enforces `workers` with polling plus `max_retries`. Workers are selected via `DATATROVE_JOB_ARRAY_INDEX`; local `logging_dir` is rejected; HF tokens/secrets are stripped from persisted `executor.pik`/`executor.json` and passed only as Job secrets. Multi-stage flows use `depends=` (fail-fast or `run_on_dependency_fail`), `HfFileSystem` cache invalidation when waiting on remote completions, and auto-creation of `hf://` logging repos/buckets.
> 
> Also exports the executor from `datatrove.executor`, documents it in the README, and adds three `*_jobs.py` examples (filter, tokenize, minhash) plus unit tests that mock the Jobs API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff0c3cc666962a20e0b41258d84f59bcbf81c0b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/filter_hf_dataset_jobs.py
+++ b/examples/filter_hf_dataset_jobs.py
@@ -15,6 +15,14 @@ Differences from the Slurm version (see comments inline):
   env_command / conda / venv to pre-activate.
 - Slurm's partition / qos / cpus_per_task / mem_per_cpu_gb / time collapse into one `flavor` (+ `timeout`).
 - `workers` caps concurrent Jobs (Slurm's `%workers`).
+
+Two gotchas the fresh-env model introduces:
+- `dependencies` must include every datatrove *extra* your steps need — not just the obvious ones.
+  Even `LambdaFilter` requires `datatrove[processing]`, because importing `datatrove.pipeline.filters`
+  eagerly imports `regex`. A missing extra fails the Job at unpickle time.
+- Point `logging_dir` at a DIFFERENT repo than your output data. If logs (`executor.json`, `stats.json`)
+  and your output parquet share one HF dataset repo, the Hub dataset viewer tries to cast them into one
+  table and errors.
 """
 
 import argparse
--- a/examples/filter_hf_dataset_jobs.py
+++ b/examples/filter_hf_dataset_jobs.py
@@ -1,5 +1,5 @@
 """
-Note: `JobsPipelineExecutor` is experimental and its API may change in future releases.
+Note: `JobsPipelineExecutor` is experimental and may change, break, or be removed in future releases.
 
 This file contains code to:
 1 - Load a parquet-format Hugging Face dataset from the hub.
--- a/examples/filter_hf_dataset_jobs.py
+++ b/examples/filter_hf_dataset_jobs.py
@@ -47,9 +47,10 @@ REMOTE_LOGS_PATH = "hf://datasets/my_org/my-filter-logs"
 # datatrove + every extra/dep your pipeline steps need, installed into each Job's uv env.
 # NB: even LambdaFilter needs the `processing` extra — importing datatrove.pipeline.filters
 # eagerly pulls in modules that require `regex`. Match the extras to the steps you use.
-# Until JobsPipelineExecutor is released, install datatrove from your branch, e.g.:
-#   "datatrove[io,processing] @ git+https://github.com/<user>/datatrove@<branch>"
-DEPENDENCIES = ["datatrove[io,processing]"]
+# Installed from git because no released datatrove ships JobsPipelineExecutor yet — with a
+# released version the Job dies at unpickle time (the class doesn't exist in the Job's env).
+# TODO: switch to "datatrove[io,processing]" once a release includes the Jobs executor.
+DEPENDENCIES = ["datatrove[io,processing] @ git+https://github.com/huggingface/datatrove"]
 
 if __name__ == "__main__":
     args = parser.parse_args()
--- a/examples/filter_hf_dataset_jobs.py
+++ b/examples/filter_hf_dataset_jobs.py
@@ -1,4 +1,6 @@
 """
+Note: `JobsPipelineExecutor` is experimental and its API may change in future releases.
+
 This file contains code to:
 1 - Load a parquet-format Hugging Face dataset from the hub.
 2 - Filter the dataset (to include only entries that contain the word 'hugging' in the text column).
--- a/examples/filter_hf_dataset_jobs.py
+++ b/examples/filter_hf_dataset_jobs.py
@@ -1,0 +1,71 @@
+"""
+This file contains code to:
+1 - Load a parquet-format Hugging Face dataset from the hub.
+2 - Filter the dataset (to include only entries that contain the word 'hugging' in the text column).
+3 - Push the filtered dataset back to the hub.
+
+...running on Hugging Face Jobs. It is the `JobsPipelineExecutor` twin of `filter_hf_dataset.py`
+(which uses Slurm) — the pipeline is identical; only the executor changes. Your machine just
+orchestrates: the filtering runs across a pool of Jobs that coordinate through a shared, *remote*
+`logging_dir`.
+
+Differences from the Slurm version (see comments inline):
+- `logging_dir` MUST be remote (hf:// or s3://) — the launcher and the Jobs are different machines.
+- `dependencies` are declared explicitly (each Job builds a fresh `uv` env); there is no
+  env_command / conda / venv to pre-activate.
+- Slurm's partition / qos / cpus_per_task / mem_per_cpu_gb / time collapse into one `flavor` (+ `timeout`).
+- `workers` caps concurrent Jobs (Slurm's `%workers`).
+"""
+
+import argparse
+
+
+parser = argparse.ArgumentParser("Filter an HF dataset on HF Jobs and push the result to the hub")
+
+parser.add_argument("input_dataset", type=str, help="HF dataset to filter")
+parser.add_argument("output_name", type=str, help="Name of the output dataset")
+parser.add_argument("--n_tasks", type=int, help="number of tasks", default=100)
+parser.add_argument("--workers", type=int, help="max concurrent Jobs (-1 = all at once)", default=20)
+parser.add_argument("--flavor", type=str, help="HF Jobs hardware flavor", default="cpu-basic")
+parser.add_argument("--text_key", type=str, help="text column", default="text")
+
+ORG_NAME = "my_org"
+# Jobs coordinate ONLY through a shared REMOTE logging_dir (hf:// or s3://), never a local path.
+# Keep this OUT of the output dataset repo — otherwise the Hub dataset viewer tries to cast
+# executor.json / stats.json into the same table as your parquet and errors.
+REMOTE_LOGS_PATH = "hf://datasets/my_org/my-filter-logs"
+# datatrove + every extra/dep your pipeline steps need, installed into each Job's uv env.
+# NB: even LambdaFilter needs the `processing` extra — importing datatrove.pipeline.filters
+# eagerly pulls in modules that require `regex`. Match the extras to the steps you use.
+# Until JobsPipelineExecutor is released, install datatrove from your branch, e.g.:
+#   "datatrove[io,processing] @ git+https://github.com/<user>/datatrove@<branch>"
+DEPENDENCIES = ["datatrove[io,processing]"]
+
+if __name__ == "__main__":
+    args = parser.parse_args()
+    from datatrove.executor import JobsPipelineExecutor
+    from datatrove.pipeline.filters import LambdaFilter
+    from datatrove.pipeline.readers import ParquetReader
+    from datatrove.pipeline.writers.huggingface import HuggingFaceDatasetWriter
+
+    dist_executor = JobsPipelineExecutor(
+        job_name=f"filter-{args.output_name}",
+        pipeline=[
+            ParquetReader(args.input_dataset, glob_pattern="**/*.parquet", text_key=args.text_key),
+            LambdaFilter(lambda doc: "hugging" in doc.text),  # add your custom filter here
+            HuggingFaceDatasetWriter(
+                dataset=f"{ORG_NAME}/{args.output_name}",
+                private=True,
+                local_working_dir=f"/tmp/{args.output_name}",  # Job-local scratch (ephemeral); cleaned up
+                output_filename="data/${rank}.parquet",
+                cleanup=True,
+            ),
+        ],
+        tasks=args.n_tasks,
+        workers=args.workers,  # concurrency cap — the analog of Slurm's `--array 0-M%workers`
+        logging_dir=f"{REMOTE_LOGS_PATH}/{args.output_name}",  # remote (hf://) — shared by launcher + Jobs
+        flavor=args.flavor,  # one hardware knob instead of partition/qos/cpus_per_task/mem_per_cpu_gb
+        dependencies=DEPENDENCIES,  # each Job builds a fresh uv env — declare deps explicitly
+        timeout="2h",
+    )
+    dist_executor.run()
--- a/examples/minhash_deduplication_jobs.py
+++ b/examples/minhash_deduplication_jobs.py
@@ -1,0 +1,132 @@
+"""
+MinHash deduplication on Hugging Face Jobs — the `JobsPipelineExecutor` twin of
+`minhash_deduplication.py` (Slurm). The 4-stage pipeline is identical; only the executors change.
+
+  stage1 signatures -> stage2 buckets (depends=1) -> stage3 cluster (depends=2) -> stage4 filter (depends=3)
+
+Notes specific to running this multi-stage dedup on Jobs (see `filter_hf_dataset_jobs.py` for the
+general Slurm->Jobs comparison):
+- `MINHASH_BASE_PATH` and the logs must be a shared REMOTE path (hf:// or s3://) that all four stages
+  read/write. HF S3-compatible buckets are a good fit and reduce read-after-write races between stages.
+- `dependencies` = `datatrove[io,processing]` PLUS `spacy`: the signature stage's English word tokenizer
+  is `spacy.blank("en")` (no model download), and `spacy` only ships in the heavy `multilingual` extra,
+  so add the bare package rather than that whole extra.
+- `max_retries >= 1` matters here: `hf://` has read-after-write lag, so a downstream stage can 404 on a
+  file the previous stage just wrote; a retry succeeds once it propagates. (Not needed on a strongly
+  consistent store.)
+- stage 1 and stage 4 MUST use the same input reader and the same task count (unchanged from Slurm).
+- Set `workers=N` on the stages to cap concurrent Jobs (the analog of Slurm's `%workers`).
+"""
+
+from datatrove.executor import JobsPipelineExecutor
+from datatrove.pipeline.dedup import MinhashDedupSignature
+from datatrove.pipeline.dedup.minhash import (
+    MinhashConfig,
+    MinhashDedupBuckets,
+    MinhashDedupCluster,
+    MinhashDedupFilter,
+)
+from datatrove.pipeline.readers import HuggingFaceDatasetReader
+from datatrove.pipeline.tokens import TokensCounter
+from datatrove.pipeline.writers.jsonl import JsonlWriter
+from datatrove.utils.hashing import HashConfig
+from datatrove.utils.typeshelper import Languages
+
+
+# you can also change ngrams or the number of buckets and their size here
+minhash_config = MinhashConfig(
+    hash_config=HashConfig(precision=64),
+    num_buckets=14,
+    hashes_per_bucket=8,
+)  # better precision -> fewer false positives (collisions)
+
+# a shared REMOTE base path all four stages use (hf:// or s3://) — never a local path
+MINHASH_BASE_PATH = "hf://datasets/my_org/my-minhash"
+LOGS_FOLDER = "hf://datasets/my_org/my-minhash-logs"
+
+# datatrove + processing (nltk/xxhash/regex/tokenizers) + bare spacy (English word tokenizer).
+# Until JobsPipelineExecutor is released, install datatrove from your branch, e.g.:
+#   "datatrove[io,processing] @ git+https://github.com/<user>/datatrove@<branch>"
+DEPENDENCIES = ["datatrove[io,processing]", "spacy"]
+
+TOTAL_TASKS = 50
+
+# this is the original data that we want to deduplicate — stage 1 and stage 4 must use the SAME reader
+INPUT_READER = HuggingFaceDatasetReader("stanfordnlp/imdb", dataset_options={"split": "train"}, text_key="text")
+
+# shared executor config for every stage
+common = {
+    "flavor": "cpu-basic",
+    "dependencies": DEPENDENCIES,
+    "timeout": "2h",
+    "max_retries": 2,  # tolerate transient hf:// read-after-write 404s between stages
+}
+
+# stage 1 computes minhash signatures for each task (each task gets a set of files)
+stage1 = JobsPipelineExecutor(
+    job_name="mh1",
+    pipeline=[
+        INPUT_READER,
+        MinhashDedupSignature(
+            output_folder=f"{MINHASH_BASE_PATH}/signatures", config=minhash_config, language=Languages.english
+        ),
+    ],
+    tasks=TOTAL_TASKS,
+    logging_dir=f"{LOGS_FOLDER}/signatures",
+    **common,
+)
+
+# stage 2 finds matches between signatures in each bucket
+stage2 = JobsPipelineExecutor(
+    job_name="mh2",
+    pipeline=[
+        MinhashDedupBuckets(
+            input_folder=f"{MINHASH_BASE_PATH}/signatures",
+            output_folder=f"{MINHASH_BASE_PATH}/buckets",
+            config=minhash_config,
+        ),
+    ],
+    tasks=minhash_config.num_buckets,
+    logging_dir=f"{LOGS_FOLDER}/buckets",
+    depends=stage1,
+    **common,
+)
+
+# stage 3 creates clusters of duplicates using the results from all buckets
+stage3 = JobsPipelineExecutor(
+    job_name="mh3",
+    pipeline=[
+        MinhashDedupCluster(
+            input_folder=f"{MINHASH_BASE_PATH}/buckets",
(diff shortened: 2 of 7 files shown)


src/datatrove/executor/jobs.py
```diff
@@ -144,6 +147,7 @@ def __init__(
         run_on_dependency_fail: bool = False,
         job_name: str = "data_processing",
     ):
+        logger.warning("JobsPipelineExecutor is experimental: its API may change (or break) in future releases.")
```
davanstrien: maybe say it may be removed?

examples/minhash_deduplication_jobs.py
```diff
@@ -10,6 +10,10 @@
 general Slurm->Jobs comparison):
 - `MINHASH_BASE_PATH` and the logs must be a shared REMOTE path (hf:// or s3://) that all four stages
   read/write. HF S3-compatible buckets are a good fit and reduce read-after-write races between stages.
+- At higher fan-out, many Jobs committing small files to a single hf:// dataset repo can hit per-repo
```
davanstrien: What about pushing to a bucket? Is this reusing existing code from datatrove?

examples/minhash_deduplication_jobs.py
```diff
@@ -10,6 +10,10 @@
...
+  commit rate limits (429 "maximum time in concurrency queue reached") in the signature stage. If you
```
davanstrien: is this with a token?

davanstrien:
## Verification log — a day of real-Jobs runs against this branch (2026-07-09)

Before undrafting I ran the executor through a battery on real Jobs — not just the happy path, but the failure and resume semantics, at a scale where infrastructure limits actually show up. Writing it up here because two of the "failures" turned out to be the most useful results of the day. ~130 cpu-basic/cpu-upgrade Jobs total; well under $10 of compute.

**The basics worked as designed.** The filter example fanned 100k imdb docs across 4 Jobs and wrote the filtered parquet back to the Hub. Tokenize ran its 16-task fan-out plus single-task `depends=` fan-in merger. The full datatrove test suite ran green inside a cpu-upgrade Job (381 passed — with `test_ray` excluded, since Ray segfaults inside Job containers regardless of this branch; GH CI covers it on normal runners). After the credential-scrub fix, I reran with an explicit `token=` and a canary secret and pulled `executor.pik`/`executor.json` back off the remote logging repo: neither credential is persisted, and the Jobs still authenticate via Job secrets.

**Failure semantics, tested with real failures.** A stage whose odd ranks deterministically raise: the coordinator saw the ERROR states, resubmitted each once (`max_retries=1`), gave up, and the dependent stage refused to run with a `RuntimeError` pointing at resume. Rerunning the same logging_dir with a fixed pipeline relaunched *exactly* the 4 previously-failed ranks — not all 8 — and the child then ran clean.

**Useful failure #1: Hub commit limits at fan-out.** Minhash at realistic scale (ag_news, 120k docs, 24 signature tasks at 8 concurrent Jobs) hit 429 "maximum time in concurrency queue reached" — 8 Jobs each committing ~14 small signature files to one `hf://` *dataset repo* saturates that repo's commit queue, and 6/24 ranks exhausted retries. The chain failed fast with resume guidance (as designed), and resuming at `workers=4` completed stage 1 by relaunching only the 6 missing ranks. The same stage with identical settings on a **Storage Bucket** (`hf://buckets/...` — same `HfFileSystem` code path, no code changes, huggingface_hub ≥ 1.6.0): **24/24, zero retries**. Bucket writes are object puts; there is no commit queue.

**Useful failure #2: buckets don't fix everything.** Stage 2's bucket-merge (a 24-way priority-queue merge over remote signature files, i.e. thousands of tiny ranged reads) runs ~45 min per task on the dataset repo and ~30 min on the bucket (14/14, zero retries) — modestly faster, but the same order of magnitude. The dominant cost is the access pattern through `HfFileSystem`, not the backend. Slurm sidesteps this with node-local disk; the fix here would be read buffering/local staging, not a store choice.

So the storage guidance that's now in the minhash example: **wide fan-out or many small files per rank → bucket; final publishable dataset → dataset repo** via `HuggingFaceDatasetWriter` (few chunky files per rank commit fine, and you get the viewer). Merge-heavy stages are slow over remote fsspec either way — a known cost, correct results.

Reference table:

| # | Scenario | Config | Result |
|---|----------|--------|--------|
| 1 | Full test suite in a Job | cpu-upgrade, `uv sync --extra testing`, `pytest ./tests/` | 381 passed, 4 skipped (`test_ray` excluded — segfaults in Job containers on any branch) |
| 2 | Single-stage fan-out (filter example) | imdb, 4 tasks / 4 workers | ✅ stats merged, output on the Hub |
| 3 | Fail-fast on failed dependency | 8 tasks, odd ranks raise, `max_retries=1`, child via `depends=` | ✅ resubmits observed → 4 ranks permanently failed → child raised `RuntimeError` |
| 4 | Idempotent resume | rerun of #3 with fixed pipeline | ✅ exactly the 4 failed ranks relaunched, child then ran |
| 5 | Tokenize→merge (example) | imdb, 16 tasks @ 8 workers + 1 merger | ✅ both stages, stats merged |
| 6 | Credential scrub | explicit `token=` + canary secret | ✅ nothing persisted to the logging_dir |
| 7 | Minhash 4-stage at scale | ag_news 120k docs, 24 tasks @ 8 workers, dataset repo | ⚠️ 6/24 ranks hit 429 commit limits → fail-fast → resume at `workers=4` completed stage 1 |
| 8 | Same signatures on a bucket | identical config, `hf://buckets` | ✅ 24/24, zero retries |
| 9 | Stage-2 merge, repo vs bucket | 14 merge tasks each | ~45 min/task (repo) vs ~30 min/task (bucket, 14/14 zero retries) — same order of magnitude; the tiny-read access pattern dominates either way |

Operational notes: per-Job overhead ≈ 1–2 min (scheduling + uv env build with `datatrove[io,processing]`) — size tasks to several minutes of work minimum. The Jobs API (`run_uv_job`/`inspect_job`) imports at exactly `huggingface-hub==1.5.0` (the current pin); `hf://buckets` paths need ≥ 1.6.0.

davanstrien:
**Follow-up: found the actual cause of the slow stage-2 merges — a one-kwarg fix, not a storage cost.**

The verification log above attributed the ~30–45 min/task merge times to "tiny ranged reads through `HfFileSystem`, backend-independent". Half right. Profiling the read path shows the culprit is that `MinhashDedupBuckets` couples the fsspec `block_size` to its unpack buffer: `lines_to_buffer (5) × line_size (68 B) = 340 bytes` (`minhash.py:420`). fsspec's readahead cache then misses every other read → one HTTP range request per ~680 bytes → ~500 GETs per signature file, × 24 open files per merge task. `HfFileSystem` itself is fine — the same file reads in a **single** request at the default 5 MB block size (micro-benchmark on one real sig file: 266 s / 500 requests → 0.63 s / 1 request).

Confirmed on real Jobs: the identical 14-task stage-2 run from the log — **65 minutes** wall at 8 workers — completes in **99 seconds** with `MinhashDedupBuckets(..., lines_to_buffer=-1)` (14/14, zero retries; wall time now dominated by per-Job env build). The example now sets `lines_to_buffer=-1`, with the full tradeoff explained in its module docstring: budget `num_stage1_tasks × sig_file_size` of RAM (~8 MB here), and at thousands of stage-1 tasks keep a bounded value and stage the bucket's sig files to the Job's ephemeral local disk instead — which is what Slurm gets for free from node-local disk.

Worth noting the tiny-block coupling is deliberate and recent (#396): with thousands of signature files open concurrently at FineWeb scale, a 5 MB fsspec buffer per stream would be tens of GB of RAM, and on Slurm's local/parallel filesystems tiny reads are essentially free — the tradeoff only turns pathological on high-latency remote filesystems, which nobody had pointed stage 2 at before this PR. `-1` has been an intended mode since #124 (stage 1 uses it internally), so the example-level setting needs no core changes.

davanstrien:
**One practical note from running a real extraction workload through the executor** (FineWeb-style pipeline over CommonCrawl WARCs, read directly from [CC's HF bucket](https://huggingface.co/buckets/commoncrawl/commoncrawl)):

Step configs tuned on dedicated Slurm cores can need adaptation on shared CPU flavors. `Trafilatura`'s default `timeout=1` is fine on an idle cpu-basic node, but under sustained load at fan-out the extraction-time tail crosses 1s en masse — and each timeout costs a sandbox kill + respawn, so jobs degrade into respawn churn. `Trafilatura(..., timeout=10)` removed the problem completely (27 timeouts across 413k pages, 0.007%).

Worth knowing when porting Slurm-tuned pipelines to Jobs. A Jobs twin of `process_common_crawl_dump.py` could be a natural follow-up.

src/datatrove/executor/jobs.py
```diff
@@ -0,0 +1,469 @@
...
+                        f"Dependency job '{self.depends.job_name}' finished with {len(incomplete)}/"
+                        f"{self.depends.world_size} tasks still incomplete. Rerun to resume, raise max_retries, "
+                        f"or set run_on_dependency_fail=True to continue anyway."
+                    )
+                # dependency was launched by another process and is still running — keep waiting
+                logger.info(f"Dependency job still has {len(incomplete)}/{self.depends.world_size} tasks. Waiting...")
+                time.sleep(2 * 60)
+            self.depends = None  # avoid pickling the dependency chain
+
+        ranks_to_run = self.get_incomplete_ranks()
```
JoelNiklaus, line 260: **Missing `logging_dir` repo → hard crash before any Job launches.**

Following any of the examples verbatim, the first thing `launch_job()` does is `get_incomplete_ranks()` → `list_files("completions")`, which raises `FileNotFoundError: datasets/<repo>/.../completions (repository not found)` when the logging repo doesn't exist yet. The *output* repo auto-creates (via `HuggingFaceDatasetWriter`), but the logging repo does not, and nothing in the examples or the docstring says to pre-create it — so every example crashes here until you manually `create_repo(...)`.

Repro: all 3 examples fail on first run; a `create_repo(logging_repo, repo_type="dataset", exist_ok=True)` beforehand fixes it and everything then runs clean.

Could `launch_job` create the repo when `logging_dir` is `hf://` (mirroring the writer), or `get_incomplete_ranks` treat a missing repo as "no completions"? At minimum worth a line in the docstring.

examples/filter_hf_dataset_jobs.py
```diff
@@ -0,0 +1,81 @@
...
+# Jobs coordinate ONLY through a shared REMOTE logging_dir (hf:// or s3://), never a local path.
+# Keep this OUT of the output dataset repo — otherwise the Hub dataset viewer tries to cast
+# executor.json / stats.json into the same table as your parquet and errors.
+REMOTE_LOGS_PATH = "hf://datasets/my_org/my-filter-logs"
+# datatrove + every extra/dep your pipeline steps need, installed into each Job's uv env.
+# NB: even LambdaFilter needs the `processing` extra — importing datatrove.pipeline.filters
+# eagerly pulls in modules that require `regex`. Match the extras to the steps you use.
+# Until JobsPipelineExecutor is released, install datatrove from your branch, e.g.:
+#   "datatrove[io,processing] @ git+https://github.com/<user>/datatrove@<branch>"
+DEPENDENCIES = ["datatrove[io,processing]"]
```
JoelNiklaus: **This default fails the Job at unpickle time (applies to all three examples).**

On a real Job, `launch_pickled_pipeline` does `dill.load(executor.pik)`, which needs `JobsPipelineExecutor` importable in the Job's env. With released `datatrove[io,processing]` the class doesn't exist yet, so the Job dies before running any rank — with an opaque unpickle error, not an obvious message. The comment right above says to point datatrove at your branch, but the actual value doesn't, so a copy-paste user gets a confusing failure.

Until release, consider making the branch URL the literal default (e.g. `"datatrove[io,processing] @ git+https://github.com/huggingface/datatrove@<branch>"` with a `# TODO: switch to released datatrove after merge`) rather than a value that contradicts its own comment. Same `DEPENDENCIES` line in `tokenize_hf_dataset_jobs.py` (L38) and `minhash_deduplication_jobs.py` (L68).

examples/minhash_deduplication_jobs.py
```diff
@@ -0,0 +1,153 @@
...
+  `block_size` (5 lines ≈ 340 bytes) — so over hf:// each couple of reads becomes its own HTTP range
+  request, ~500 requests per signature file (measured: this stage went 65 min → 99 s after switching
+  to -1). -1 instead reads each file in a single request but buffers whole files in RAM: budget
+  `num_stage1_tasks × sig_file_size` (here 24 × ~340 KB ≈ 8 MB — trivial). At thousands of stage-1
+  tasks, keep a bounded `lines_to_buffer` and stage the bucket's signature files to the Job's local
+  ephemeral disk first instead.
+- `dependencies` = `datatrove[io,processing]` PLUS `spacy`: the signature stage's English word tokenizer
+  is `spacy.blank("en")` (no model download), and `spacy` only ships in the heavy `multilingual` extra,
+  so add the bare package rather than that whole extra.
+- `max_retries >= 1` matters here: `hf://` has read-after-write lag, so a downstream stage can 404 on a
```
JoelNiklaus: Confirmed on a real run — and it's a touch worse than "a downstream stage can 404": even stage 1's **own** post-write self-verification read in `MinhashDedupSignature` (reopening the `.sig` file it just wrote, same task, milliseconds apart) 404'd with `RemoteEntryNotFoundError` on an `hf://` dataset repo. `max_retries` recovered it (resubmit succeeded), but stage 1 also crawled (~6 min for 6.25k docs) from the per-file write+verify round-trips.

Strong argument for defaulting these examples to `hf://buckets/...` rather than dataset repos, as you already recommend in the docstring — the dataset-repo path is functional but noticeably slow + flaky here.

JoelNiklaus (commented):
Took this branch for a spin: ran all three Jobs examples as **real** HF Jobs (`filter` single-stage, `tokenize` 2-stage `depends=`, `minhash` 4-stage `depends=`) plus the unit suite (11/11 pass), on `cpu-basic`, ~$0.02 total.

The executor itself works end-to-end — `depends=` chaining, the `workers` concurrency cap, per-rank stats merge, credential scrub, and the retry loop all behaved correctly (one stage-1 Job genuinely failed mid-run and was auto-resubmitted `1/2` and completed). Nice work.

Three things a fresh user hits when following the examples verbatim, left as inline comments below. Happy to open a PR for any of them.

davanstrien:
Thanks for the review @JoelNiklaus — having all three examples run end-to-end independently is exactly the verification this needed. All three points addressed:

- ea0661d: `launch_job` now auto-creates the `hf://` logging repo/bucket (private, `exist_ok=True`, mirroring the writers), with tests
- 18ae3f4: the examples' `DEPENDENCIES` now literally install from git, and the minhash example defaults to a bucket — folded your same-task `.sig` 404 observation into the docstring note

One deliberate difference from your suggestion on the install URL: it points at upstream main rather than my branch, since the examples ship with the merge (at which point main has the executor and a fork branch goes stale). There's a TODO to drop the git pin entirely after the next release.

julien-c:
great work @davanstrien!

PR #505, opened 2026-08-07
ci: switch PyPI release to trusted publishing (OIDC)

davanstrien:
The May 6 release attempt passed tests and failed at the "Publish to test PyPI" step on PyPI keys. Rather than rotate the tokens, this switches publishing to [trusted publishing](https://docs.pypi.org/trusted-publishers/): GitHub proves the workflow's identity to PyPI per-run via OIDC, and no long-lived secret exists to expire again.

**Changes**
- Upload step → `pypa/gh-action-pypi-publish` (SHA-pinned, v1.14.1 — same pin as [lerobot's release workflow](https://github.com/huggingface/lerobot/blob/ef88d4e52b9f3a16638e4b73202d619b7606fd41/.github/workflows/release.yml#L113))
- `id-token: write` added to the release job's permissions; `TWINE_USERNAME` env and the twine install dropped
- **TestPyPI removed from the flow.** 
  - the install-back gate now installs the **built wheel from `dist/`** and runs the full test suite *before* publish — same artifact check, no scratch registry, and it deletes the fragile curl/grep of the TestPyPI simple index (which matched the raw pyproject version string against PEP 440-normalized filenames)
  - the registry-round-trip check moves to the rc rehearsal (below)
- **Tag creation moved after the publish** — a failed upload no longer leaves a stale `vX.Y.Z` tag

Resulting order: tests → build → **install wheel + pytest** → publish to PyPI → tag. Everything before publish can fail and re-run freely; "version exists" on PyPI stays a hard failure. `TEST_PYPI_TOKEN` and `PYPI_TOKEN` become unused on merge and can be deleted from repo secrets.

**Tested — dry run on this branch ([run 31608389848](https://github.com/huggingface/datatrove/actions/runs/31608389848))**

The trusted publisher is configured on pypi.org (owner `huggingface`, repo `datatrove`, workflow `pypi-release.yml`, no environment) and the workflow was dispatched on this branch as-is — version still 0.9.0, so nothing could publish. Result, every step proven except the final upload itself:
- test suite green (3.10–3.13), build green
- wheel-install gate green: the built wheel installs with `[testing]` extras and passes the full suite
- **OIDC exchange succeeded** against the live publisher config (log shows `username: __token__` + sigstore attestations) — the run reached a real authenticated upload attempt
- PyPI rejected with `400 File already exists (datatrove-0.9.0...)`, exactly as intended: existing release files are immutable, and no tag was created (tag runs after publish)

The only untested path is a successful upload + tag, which is what the `0.10.0rc1` rehearsal (#506) exercises after merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Changes the production release and tagging path and removes TestPyPI as a safety net; misconfigured trusted publishing or workflow permissions could block or mis-route releases until fixed.
> 
> **Overview**
> **Replaces long-lived PyPI/TestPyPI twine tokens** with GitHub OIDC trusted publishing via `pypa/gh-action-pypi-publish`, adding `id-token: write` and dropping twine/`TWINE_*` usage.
> 
> **Release flow is reordered and simplified:** after build, the workflow installs the **local wheel** from `dist/` (with `[testing]`) and runs the full pytest suite, then publishes to production PyPI, then creates the git tag—so a failed upload no longer leaves a tag and TestPyPI is removed entirely.
> 
> **Pre-publish validation** no longer depends on scraping the TestPyPI simple index; the same “install what we ship” check happens against the artifact about to be uploaded.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 625d5c177b3a939ce2fd6cb14c625acff2a7d32e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/pypi-release.yml
+++ b/.github/workflows/pypi-release.yml
@@ -31,11 +31,19 @@ jobs:
       - name: Build the dist files
         run: python -m build .
 
-      - name: Publish to the test PyPI
+      - name: Test installing the built wheel and running tests
+        # same gate the old install-from-TestPyPI step provided, without the
+        # scratch registry: prove the built artifact (not the source tree)
+        # installs and passes tests before anything irreversible happens
+        run: |
+          python -m pip install uv
+          uv pip install --system "$(ls dist/*.whl)[testing]"
+          python -m nltk.downloader punkt
+          python -m pytest -sv ./tests/
+
+      - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@ba38be9e461d3875417946c167d0b5f3d385a247  # v1.14.1
         with:
-          repository-url: https://test.pypi.org/legacy/
-          skip-existing: true
           verbose: true
           print-hash: true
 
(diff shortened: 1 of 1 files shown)


davanstrien:
The Secret Leaks failure is an upstream trufflehog bug, not this PR: the Lob detector is currently false-positive-verifying snake_case identifiers (trufflesecurity/trufflehog#5184, trufflesecurity/trufflehog#5187). Reproduced locally — the "secrets" are pytest function names like `test_zero_shows_waiting_for_first_upload`. The workflow pulls `trufflehog:latest`, so a re-run after the upstream fix should go green.

PR #517, opened 2026-08-29
feat(inference): add OrcaRouter as a first-class inference server

putraperdana1207-pixel:
Adds [OrcaRouter](https://www.orcarouter.ai) as a first-class `server_type` for the `InferenceRunner`, so datatrove users who run synthetic-data generation through the block can point their rollout jobs at OrcaRouter's gateway with a single `server_type="orcarouter"` instead of configuring an anonymous custom base URL.

OrcaRouter is an OpenAI-compatible AI gateway built for both models and agents. Like the OpenAI-compatible HTTPS endpoints datatrove already supports, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding `orcarouter` as a named provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

The implementation mirrors the existing `EndpointServer` (same OpenAI client, same retry/error mapping, same localhost fast path for tests) and adds an `is_ready` probe that accounts for the provider base URL already carrying the `/v1` prefix. The API key is read from `InferenceConfig.api_key` or the `ORCAROUTER_API_KEY` environment variable, and `endpoint_url` can be overridden for self-hosted gateways.

Verified with a live request through `InferenceRunner` (`server_type="orcarouter"`, `model_name_or_path="orcarouter/auto"`), which routed to a real model and returned a completion. Added unit tests (mock-endpoint round trip and missing-key validation); `make quality` and the inference test suite pass.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

I'm an engineer on the OrcaRouter team.

---

> [!NOTE]
> **Low Risk**
> Changes follow the existing external `EndpointServer` pattern and only add optional outbound API calls; no changes to core auth or data handling in datatrove itself.
> 
> **Overview**
> Adds **OrcaRouter** as a named `server_type="orcarouter"` on `InferenceConfig` and wires it through `InferenceRunner` server factory and exports, so synthetic-data rollouts can target OrcaRouter without treating it as a generic `endpoint`.
> 
> Introduces `OrcaRouterServer`: defaults to `https://api.orcarouter.ai/v1`, requires `api_key` or `ORCAROUTER_API_KEY`, allows `endpoint_url` override for self-hosted gateways, probes readiness via `/models` or `/v1/models` depending on whether the base URL already ends with `/v1`, and uses the same OpenAI async client + error/retry mapping as `EndpointServer` (with a localhost HTTP fast path for tests).
> 
> README lists OrcaRouter among supported backends; tests cover a mock localhost round trip and missing API key validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c78b1b2db1d5e49f6151f513665476e8fcd6f6be. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


davanstrien:
Thanks for the PR! OpenAI-compatible gateways like OrcaRouter are already supported via `server_type="endpoint"` with a custom `endpoint_url` and `api_key`, so we'd rather not add vendor-specific server types as first-class options. The `/v1` readiness issue you worked around is addressed in #518

PR #519, opened 2026-09-04
fix(hashing): support xxhash 4

### LINKED ISSUE #508: xxhash is pinned to <4: the proper fix needs a decision on str-vs-bytes hashing
`xxhash` is currently pinned to `<4` in the `processing` extra (#507). This issue records why, and why lifting the pin is not just a one-line change.

## What broke

[xxhash 4.0.0](https://pypi.org/project/xxhash/4.0.0/) (2026-08-12) removed implicit `str` encoding. `xxh32_intdigest` / `xxh64_intdigest` now raise `TypeError: Strings must be encoded before hashing` for `str` input.

`datatrove.utils.hashes.xxhash` passes `str` straight through, so with `xxhash>=4` every dedup pipeline fails: minhash, exact dedup, URL dedup, sentence dedup, bloom filter. Lockfiles hid this from CI; it surfaced in the `0.10.0rc1` release run, at the step that installs the built wheel and resolves dependencies fresh.

`datasets` also declares `xxhash` unbounded, but hashes bytes, so it is unaffected.

## Current state

`xxhash<4` in the `processing` extra, shipped in `0.10.0rc1`. This restores exactly the dependency set everything already had, with no behaviour change. It is a stopgap: an upper bound on a common dependency will eventually cause resolution conflicts for downstream users.

## Why the obvious fix needs thought

The obvious fix is to encode inside the wrapper:

```python
def xxhash64(data: str | bytes):
    if isinstance(data, str):
        data = data.encode("utf-8")
    return xxhash.xxh64_intdigest(data)
```

UTF-8 is what xxhash `<4` encoded implicitly, so digests are unchanged and existing signatures stay comparable (verified against 4.0.0). But it costs measurably on the hottest call in the library — minhash hashes every shingle of every document. Micro-benchmark, 300k calls on a typical shingle:

| path | ns/call |
| --- | --- |
| xxhash 3.x, `str` in (current behaviour) | 72 |
| Python-side `.encode()` on every call | 102 |
| ... plus the `isinstance` guard | 122 |
| already-`bytes` input | 62 |

xxhash 3.x encoded in C using CPython's cached UTF-8 representation of the `str`. A Python-level encode allocates a new `bytes` object per call, which is why the fix is ~1.4x, and the guarded version ~1.7x. (Micro-benchmark only — I have not measured end-to-end impact on a real dedup run.)

The last row is the interesting one: hashing `bytes` is *faster* than what we do today. So the better fix may be to hash bytes end-to-end in the dedup steps rather than to encode inside the wrapper — turning a forced regression into a small win.

## A wrinkle for any bytes-based approach

`create_hash_func(config, input_type)` already dispatches str/bytes for sha1 (`sha1_hash64` vs `sha1_hash64_bytes`), but xxhash has a single pair that accepted both. The only caller passing a non-default `input_type` is `ExactDedup`, which derives it by reflection:

```python
hash_fc_type = inspect.signature(self.config.content_getter).return_annotation
```

If the user's `content_getter` is unannotated, or is defined in a module using `from __future__ import annotations` (where the annotation arrives as the string `"bytes"`, not the type), this does not resolve to `bytes` and the `str` function is returned. Today xxhash tolerates that mismatch because it accepted both types; a strict bytes/str split would surface it as a crash. Worth handling deliberately rather than discovering it downstream.

## Options

1. Encode in the wrapper, lift the pin. Smallest diff, ~1.4x on the hash call.
2. Hash bytes end-to-end in the dedup steps, lift the pin. Larger change, likely a net speedup, needs the reflection wrinkle above resolved.
3. Keep the pin for now. Zero risk, no user-visible cost today, revisit when something in the ecosystem actually needs `xxhash>=4`.

Option 1 is written and digest-verified if a quick unblock is ever wanted. Option 2 looks like the right long-term shape but deserves benchmarks on a real dedup run rather than a micro-benchmark.

tadad:
## Summary

- Encode `str` inputs as UTF-8 before calling the xxhash 32-bit and 64-bit one-shot functions.
- Preserve direct `bytes` hashing and the digest values produced by xxhash 3.
- Remove the temporary `xxhash<4` dependency bound.
- Refresh `uv.lock` so the locked development environment resolves xxhash 4.0.1.
- Add regression coverage for non-ASCII text and both supported hash widths.

Fixes https://github.com/huggingface/datatrove/issues/508

## Approach

This implements Option 1 from the issue. xxhash 4 rejects `str` inputs, so the DataTrove wrapper now encodes strings before calling `xxh32_intdigest` or `xxh64_intdigest`. Byte inputs pass through unchanged.

Keeping the conversion in the wrapper preserves DataTrove's existing `str | bytes` interface and avoids the broader `ExactDedup` type-dispatch changes required by a bytes-only pipeline. The tradeoff is that UTF-8 encoding remains part of each hash call. The issue's benchmark measured this option at about 1.4 times the hash-call cost of passing pre-encoded bytes.

## Validation

- `uv lock --check`
- `uv tree --locked --package xxhash` resolves xxhash 4.0.1
- Locked hashing and deduplication suite: 28 passed, 3 skipped
- Deduplication, hashing, and request-cache tests with `xxhash==3.8.1`: 29 passed, 3 skipped
- Deduplication, hashing, and request-cache tests with `xxhash==4.0.1`: 29 passed, 3 skipped
- Cross-version digest comparison for empty, ASCII, Unicode, embedded-NUL, and large inputs across xxh32, xxh64, and xxh128
- `uv run --isolated --locked --extra quality make quality`
- Branch-level autoreview: no actionable findings

The three skipped tests are the existing n-gram decontamination cases disabled because Lighteval does not support NumPy 2.

---

> [!NOTE]
> **Medium Risk**
> Hashing feeds deduplication and related processing; incorrect string encoding would change fingerprints, though the change is narrow and covered by new regression tests.
> 
> **Overview**
> **xxhash 4** no longer accepts `str` inputs, so this PR makes DataTrove compatible by **UTF-8 encoding strings** in `xxhash32` / `xxhash64` before calling the library, while **leaving `bytes` unchanged** so digests stay aligned with xxhash 3 behavior.
> 
> The **`xxhash<4` pin is removed** in `pyproject.toml` and **`uv.lock` is refreshed** to **xxhash 4.0.1**. Return types are annotated on the wrapper helpers, and a regression test checks that hashing `"héllo"` matches its UTF-8 bytes for both 32- and 64-bit widths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 11269bd29b6213e311bec3506eb2b9bafacf8087. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tadad:
I also implemented and benchmarked Option 2 from #508, which moves the deduplication pipelines to byte-oriented hashing. On xxhash 4.0.1, it reduced an isolated byte-hash call from 66.8 ns to 52.4 ns, a 21.5% improvement. That gain did not translate into a meaningful end-to-end difference: across controlled 1,700-word shingling workloads, results ranged from 4.6% faster to 0.5% slower, and the test using DataTrove’s English tokenizer and representative minhash fixture improved from 3.457 ms to 3.445 ms per document, or 0.35%, which is within benchmark noise. Option 2 also materializes encoded token lists, giving it a potentially higher peak-memory cost for large documents, and requires coordinated changes across the deduplication implementations. Although it creates a stricter bytes boundary and removes `ExactDedup`’s reliance on return annotations, the benchmarks do not show a user-noticeable performance benefit. Given that result, Option 1 appears to be the better tradeoff: it preserves the existing interface and digest values with a substantially simpler implementation.

lpla:
Hiya. I coincidentally independently benchmarked #508 while this PR was being prepared, using complete DataTrove deduplication pipelines rather than only the hash call. My results agree with the direction drafted here. I don't see enough end-to-end benefit to justify moving the dedup pipelines to bytes.

Across MinHash, Exact, URL, Sentence and Bloom on a 2,200-document FineWeb-derived workload (~1.1M words), the UTF-8 wrapper was effectively indistinguishable from the existing 3.x path at pipeline level. For MinHash, 4.0.1 + wrapper was 18.008 s median versus 17.985 s on 3.8.1; byte-oriented prototypes did not show a statistically clear improvement. I also repeated MinHash against the lockfile's xxhash 3.6.0 with the same conclusion.

A larger 44k-document ExactDedup workload did show a modest ~2.7% improvement from hashing pre-encoded bytes, but I don't think that justifies a broad refactor given the lack of consistent benefit elsewhere and the extra memory.

I also verified cross-version compatibility, like digest vectors, dedup signatures/intermediate decisions/final output, and consuming indexes produced with xxhash 3.6.0 from 4.0.1 all remained compatible in the configurations I tested.

One separate issue did fall out of the investigation though. ExactDedupSignature currently selects the hash function by inspecting content_getter's return annotation. Unannotated getters, postponed "bytes" annotations and some uninspectable callables therefore select (or) fail incorrectly; for SHA1 this bug exists independently of xxhash 4.

I have a small and drafty prototype that instead normalizes the actual getter result (str -> strict UTF-8, bytes unchanged), calls the getter only once, and hashes bytes. It preserves the existing digests in my compatibility tests.

If that sounds worthwhile, I can open a separate focused issue for the content_getter behavior rather than mixing it into this compatibility PR. I also have the full benchmark harness and results available in Markdown, if useful.

PR #522, opened 2026-09-11
fix(extractors): honor PEP 508 pip names and empty sandbox warmup

### LINKED ISSUE #521: ReadabilityInscriptis fails dependency validation and the sandbox's empty warmup
I found two initialization problems while testing the existing `ReadabilityInscriptis` extractor.

With the documented Hugging Face readability fork installed, this still raises `ImportError`:

```python
import importlib.metadata
from datatrove.pipeline.extractors import ReadabilityInscriptis

print(importlib.metadata.version("readability-lxml"))  # 0.8.3.dev0
ReadabilityInscriptis(timeout=2)
```

The dependency declaration contains the full direct reference (`readability-lxml @ git+https://github.com/huggingface/python-readability.git@speedup`), but `_is_distribution_available` compares it literally with the installed distribution's `Name`, `readability-lxml`.

After overriding that declaration locally to isolate the next step, extraction from nonempty HTML succeeds directly. However, `extract("")` raises `Unparseable: Document is empty`. `BaseExtractor.run` sends that empty string as its sandbox warmup, so the worker fails before processing the real document. My one-document reproduction yields no documents and records a warmup timeout.

cpruijsen:
## Summary

`_is_distribution_available` now compares the PEP 508 name (`readability-lxml`) rather than the full `name @ git+...` specifier. The `ImportError` still prints the original pip argument, so the fork install hint is kept. `ExtractorSandbox` treats warmup failures as non-fatal (warmup is only there to import extractor code) and still sends the ready signal. `ReadabilityInscriptis.extract("")` returns `""` after those imports, instead of raising `Unparseable: Document is empty`.

Fixes #521.

`ReadabilityInscriptis` could not be constructed when the documented Hugging Face `readability-lxml` fork was installed, and a one-document `run()` yielded nothing because the extractor sandbox died on its empty warmup. The reporter reproduced the `ImportError` with `readability-lxml==0.8.3.dev0`, `extract("")` raising `Unparseable: Document is empty`, and a warmup timeout that dropped the only document. All three still happened on this tree at HEAD before the change. The existing `test_basic_article_readability` also could not construct the class with that fork.

## What I chose and the alternative

Parse the PEP 508 name rather than changing the dependency tuple to `("readability", "readability-lxml")`. Alternative: drop the git URL. The second tuple element is the pip-install argument everywhere else; keeping it preserves the fork hint.

Swallow warmup exceptions rather than changing default `warmup_text` to HTML. Alternative: a tiny HTML warmup document. Empty warmup is the original sandbox contract; short HTML can still fail readability's `min_text_length` / `min_text_score`.

Return `""` for empty HTML rather than leaving `extract("")` raising. Alternative: sandbox-only fix. Trafilatura already returns empty instead of raising; `run` already drops empty extracted text.

Can use the short pip name, an HTML warmup, or drop the empty-input guard.

## Test plan

- [x] `pytest -sv ./tests/pipeline/test_base.py ./tests/pipeline/stats/test_extractors.py`
- [x] PEP 508 tests fail without the import-utils change and pass with it (numpy is a hard dependency)
- [x] Dummy extractor that raises on `""` fails warmup without the sandbox change and processes a real document with it
- [x] With the Hugging Face readability fork installed: `ReadabilityInscriptis(timeout=2)` constructs, `extract("") == ""`, and `run()` forwards a nonempty HTML document
- [ ] CI: readability tests remain skipped unless `readability-lxml` is installed; PEP 508 and dummy-sandbox tests should run on the testing extra

---

> [!NOTE]
> **Low Risk**
> Targeted fixes to dependency parsing, sandbox warmup, and empty-input handling in extractors; no auth or data-path changes.
> 
> **Overview**
> Fixes **ReadabilityInscriptis** and the extractor sandbox when dependencies are declared as PEP 508 direct URLs (e.g. the Hugging Face `readability-lxml` fork) and when warmup uses an empty document.
> 
> **Dependency checks** now normalize pip requirement strings before comparing to installed distributions—stripping `@ git+...`, version specifiers, and extras—while **ImportError** messages still show the full original `pip install` argument.
> 
> **ExtractorSandbox** no longer treats warmup failures as fatal; warmup only forces imports, so an empty default warmup that raises (e.g. readability *Document is empty*) still signals ready.
> 
> **ReadabilityInscriptis.extract** returns `""` for empty input instead of calling readability on empty HTML.
> 
> Tests cover PEP 508 distribution matching, empty HTML extraction, sandbox warmup with a raising extractor, and a full `run()` after empty warmup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 40fe077f8330ed53170c25ae77977dd5ba68a46d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lpla:
Thanks for picking up #521. I reran my original reproduction against `40fe077` on macOS / Python 3.12.12, with the actual Hugging Face `speedup` fork at `0ca5bbbb` (`readability-lxml==0.8.3.dev0`), without overrides on the PR code.

- Construction succeeds and `extract("")` returns `""`.
- `run()` forwards the original nonempty fixture without a warmup timeout, with exactly the same text as direct extraction.
- An empty document followed by nonempty HTML and a Unicode control drops only the empty document; subsequent outputs still match direct extraction.

`tests/pipeline/test_base.py` and `tests/pipeline/stats/test_extractors.py`: **10 passed, no skips**. Running those same tests against the base source gives 6 failures and 4 passes. I also reproduced the original constructor failure and, after isolating the dependency check as in #521, the empty-warmup failure on the base.

This resolves both failures I reported in #521 in this environment.

PR #527, opened 2026-09-17
Add configurable DATATROVE_LOG_LEVEL env var (#375)

### LINKED ISSUE #375: Set log level
Hi!

I was wondering it is possible to set the (verbosity) log level on Datatrove?

More specifically, I want to block/disable logs such as [this](https://github.com/huggingface/datatrove/blob/370d23a69d1fe09b467f5a5125879649edbc8d79/src/datatrove/utils/logging.py#L90).

AmulyaP07-15:
Closes #375.

Adds a `DATATROVE_LOG_LEVEL` environment variable so users can control console log verbosity, mirroring the existing `DATATROVE_COLORIZE_LOGS` pattern in `src/datatrove/utils/logging.py`.

- Defaults to `INFO`, so existing behavior is unchanged.
- Applied to both `setup_default_logger` and the master-rank console handler in `add_task_logger`; non-master ranks stay at `ERROR` as before.
- Example: `DATATROVE_LOG_LEVEL=WARNING python myscript.py` shows only warnings and errors.

Added tests in `tests/utils/test_logging.py` covering the default (INFO shown) and a raised level (INFO hidden, WARNING shown).

---

> [!NOTE]
> **Low Risk**
> Observability-only change with INFO default preserving prior behavior; no auth, data, or pipeline logic touched.
> 
> **Overview**
> Adds a **`DATATROVE_LOG_LEVEL`** environment variable (default **`INFO`**, normalized to uppercase) so console verbosity is configurable, following the same pattern as the existing colorize env vars.
> 
> The level replaces the hardcoded **`INFO`** on stderr for the default logger and for the master rank (`local_rank == 0`) in **`add_task_logger`**; worker ranks still log to stderr at **`ERROR`**, and per-task log files remain at **`DEBUG`**.
> 
> New subprocess-based tests in **`tests/utils/test_logging.py`** assert that unset env shows INFO+WARNING and **`DATATROVE_LOG_LEVEL=WARNING`** hides INFO while keeping WARNING.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6ef29573bf22db839f77c679b72d0a913758ec2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


lpla:
One default-behavior detail at `6ef2957`: with both `DATATROVE_LOG_LEVEL` and `LOGURU_LEVEL` unset, the existing `setup_default_logger()` emits DEBUG messages, whereas this PR defaults it to INFO. This also changes the logger restored by `close_task_logger()`.

Minimal comparison on the base and PR:

```sh
env -u DATATROVE_LOG_LEVEL -u LOGURU_LEVEL python -c 'from datatrove.utils.logging import logger; logger.debug("debug probe")'
```

The base prints the probe; the PR does not. The new default-level test only checks INFO/WARNING, so it does not detect this difference. I also checked the task lifecycle: task files still receive DEBUG, and secondary-rank console handlers remain at ERROR as intended.

Could we either retain the previous default for the general logger or document this as an intentional default change and add a DEBUG-level regression test? The PR description currently says existing behavior is unchanged, which is accurate for the default master-rank task handler but not the general logger.

PR #528, opened 2026-09-22
fix(filters): GopherRepetitionFilter raises ZeroDivisionError on empty documents

Linxiushen:
## Problem

`GopherRepetitionFilter.filter()` in `src/datatrove/pipeline/filters/gopher_repetition_filter.py` computes ratios by dividing by the document's word count and character count without guarding the empty case. A document whose `text` is empty therefore raises

```
ZeroDivisionError
```

at the first ratio computation, instead of being filtered out. `FineWebQualityFilter` handles the same situation by rejecting empty documents up front; this filter does not.

Reachability, stated plainly: the default reader path skips empty text on read (`readers/base.py` returns `None` when `text` is falsy), so the standard pipeline usually does not reach it. But `GopherRepetitionFilter` is a public filter that can be applied to documents from any source (custom readers, an upstream step that emits an empty `text`, in-memory documents), and there it crashes the run.

## Fix

Return `(False, "empty")` before any division when the text is empty. This mirrors `FineWebQualityFilter` and is the correct decision: an empty document should be dropped. A per-division guard would instead let an empty document pass every check and be *kept*, which is wrong.

## Test

`test_gopher_repetition_rejects_empty_doc` in `tests/pipeline/filters/test_filters.py` fails on `main` with `ZeroDivisionError` and passes with the fix. A value differential over 9 input classes (whitespace, duplicate paragraphs/lines, n-gram paths, clean text) shows the only changed output is the empty-document case; the other eight are byte-identical.

`ruff check` and `ruff format --check` pass at the pinned 0.11.7. The test needs no `nltk`/`spacy` because the empty path returns before any tokenizer runs.

---
This fix was developed with AI assistance (Claude); the change and the test were reviewed and verified locally before submission.

---

> [!NOTE]
> **Low Risk**
> Small guard on an edge case before existing logic; behavior for non-empty documents is unchanged.
> 
> **Overview**
> **`GopherRepetitionFilter`** now rejects documents with empty `text` up front with `(False, "empty")`, instead of crashing with **`ZeroDivisionError`** when repetition ratios divide by `len(text)`.
> 
> This matches the empty-document behavior of **`FineWebQualityFilter`** and ensures empty docs are dropped rather than kept if only per-check guards were added.
> 
> A unit test **`test_gopher_repetition_rejects_empty_doc`** asserts the `"empty"` rejection reason for `get_doc("")`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f119b88fa37ec6db15d0215763ef23701c70c1c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


davanstrien:
thanks!

